PR #19, opened 2020-05-03
LaTex -> markdown WabiSabi v0.2

DanGould:
Source is in README.tex.md ported from the overleaf.

The [TeXify](https://github.com/apps/texify) plugin is enabled on my fork, so every time I push a new *.tex.md file on any branch, it generates a new *.md file with the Latex rendered as SVGs using [readme2tex](https://github.com/leegao/readme2tex), which you may install locally to test what the generated images look like. Or just install the Texify app it on a fork of your own and create a pr when you're satisfied


nopara73:
You should not push autogenerated things to GitHub. GitHub is for hosting source.

PR #21, opened 2020-05-03
[WIP] Markdown conversion

nothingmuch:
Automated version of #19 using pandoc and readme2tex.
 
- [ ] pandoc is first used to convert latex to markdown, leaving math as is
  - [x] \maketitle
  - [ ] abstract - YAML metadata output with --standalone, kind of ugly
  - [ ] citations
- [x] readme2tex is then used to convert the math to SVGs
  - [x] indented math (e.g. in lists) is not converted, as a workaround
        the only enumeration was just flattened to paragraphs.
  - [x] inline math following display math in the same paragraph is
        rendered as display math, used \noindent as a workaround
- [ ] choose output file, currently overwrites README.md, if we go with this we should integrate the existing content into the latex, if not a separate output file should be choosen
- [ ] git post commit hook
- [ ] upstream latex workarounds to overleaf

Tested using debian 10's pandoc & texlive, with readme2tex from pypy

Depends on #20


nopara73:
You should not push autogenerated things to GitHub. GitHub is for hosting source.

PR #28, opened 2020-05-16
abstract & introduction

nothingmuch:
rewrote the existing opening paragraphs to be more in line with standard
conventions, and establish the background and problem statement more
clearly.

### DIFF
--- a/main.tex
+++ b/main.tex
@@ -17,35 +17,33 @@
 \maketitle
 
 \begin{abstract}
+  Bitcoin's model for value transfer is a publicly verifiable ledger of transactions, with ownership of coins defined in terms of public keys. Although this design lacks strong guarantees for privacy, it does not rule it out.
+  Despite the potential for private use, research has shown that users' activity can often be traced in practice.
+  This lack of inherent privacy has given rise to business models built on dragnet surveillance of Bitcoin users which is problematic for Bitcoin's fungibility.
 
-% TODO citations should not be included in the abstract, the abstract should be cleaned up, and the remaining information and citations split off into an introduction section with citations
+  A number of methods have been proposed to mitigate these issues. Among these, CoinJoin is an approach to structuring transactions which adds ambiguity and breaks common assumptions that underlie heuristics used for surveillance.
+  Current implementations of CoinJoin transactions present a number of limitations which may partly explain their lack of widespread adoption.
 
-Chaumian CoinJoin~\cite{mizrahi2013blind}\cite{maxwell2013coinjoin} is a technique to facilitate untrusted construction of collaborative Bitcoin transactions, also known as CoinJoins, by utilizing Chaumian blind signatures~\cite{chaum1983blind}. However this technique requires standard denominations, which limits how such transactions can be constructed.
-
-We propose to switch to a Keyed-Verification Anonymous Credentials-based (KVAC) scheme~\cite{chase2019signal} to enable more flexible transaction styles, like SharedCoin and CashFusion~\cite{cashfusion} style transactions and Knapsack~\cite{maurer2017anonymous} mixing. Our generalization also enables consolidation of UTXOs, minimizing unmixed change, relaxing minimum required denominations, payments in CoinJoins, better block space efficiency, and PayJoins in CoinJoins. We call this new protocol: WabiSabi.
-
+  This work introduces a new protocol for centrally coordinated CoinJoin implementations utilizing keyed verification anonymous credentials and homomorphic value commitments. This is an improvement on the state of the art in both privacy and flexibility, enabling additional use cases and reduced overheads.
 \end{abstract}
 
-% Important remainder from removed Motivation: In the current protocol input registration links the user's inputs, as well as their change output.
+\section{Introduction}
 
-% original motivation section had is several things rolled into one, which should be clarified separately:
-% 1. payments from a coinjoin
-% 2. payments in a coinjoin, where the amount does not appear on the blockchain
-% 3. toxic change
+Bitcoin transactions redistribute funds by consuming as inputs the outputs of previous transactions, and creating new outputs. The protocol rules enforced by the network ensure that transactions do not arbitrarily inflate the monetary supply and that output are spent at most once. While some newer cryptocurrencies use more sophisticated approaches to defining such rules, in Bitcoin the amounts as well as the specific outputs being spent are broadcast in the clear as part of the transaction. This presents significant challenges for transacting with privacy\footnote{In this work we restrict the discussion of Bitcoin privacy to that of transactions, but there are other considerations especially at the network layer. For a more comprehensive discussion see \url{https://en.bitcoin.it/wiki/Privacy}.} as shown already in some of the earliest academic studies of Bitcoin~\cite{ron2013quantitative,meiklejohn2013fistful,moser2013anonymity}.
 
-% brief explanation of why coinjoins are desirable:
-% - privacy
-% - fungibility
-% - theoretically transaction efficiency
-% and overview of where zerolink/wasabi currently falls short (variable but constrained denomination, only send to self, linkage of inputs during registration,
+The conditions for spending an output are specified in its \texttt{scriptPubKey}, typically requiring that the spending transaction be signed by a specific public key. The signatures authorizing a transaction usually commit to the transaction in its entirety, which makes it possible for mutually distrusting parties to jointly create transactions without risking misallocation of funds: participants will only sign a proposed transaction after confirming that their desired outputs are included.
 
-\section{Protocol Overview}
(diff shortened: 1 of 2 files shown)


main.tex
```diff
@@ -15,35 +15,33 @@
 \maketitle
 
 \begin{abstract}
+  Bitcoin's design is based on pseudonymous identities, and a publicly verifiable ledger of transactions transferring value between them. Although this design lacks strong guarantees for privacy, it does not rule it out.
```
nopara73, line 18: > Bitcoin's design is based on pseudonymous identities

IMO this is a vast oversimplification, and doesn't stand for scrutiny.

main.tex
```diff
@@ -15,35 +15,33 @@
...
+  Despite the potential for private use, research has shown that users' activity can often be traced in practice.
+  This lack of inherent privacy has given rise to business models built on dragnet surveillance of Bitcoin users which is problematic for Bitcoin's fungibility.
 
-% TODO citations should not be included in the abstract, the abstract should be cleaned up, and the remaining information and citations split off into an introduction section with citations
+  A number of methods have been proposed to mitigate these issues. Among these, CoinJoin is an approach to structuring transactions which adds ambiguity and breaks some common assumptions that underlie heuristics used for surveillance.
```
nopara73, line 22:
```suggestion
  A number of methods have been proposed to mitigate these issues. Among these, CoinJoin is an approach to structuring transactions which adds ambiguity and breaks some assumptions those underlie heuristics used for surveillance.
```

main.tex
```diff
@@ -15,35 +15,33 @@
...
+  Current implementations of CoinJoin transactions present a number of limitations which may partly explain their lack of widespread adoption.
```
nopara73, line 23:
```suggestion
  Current implementations of CoinJoin transactions present a number of limitations which partly explains their lack of widespread adoption.
```

main.tex
```diff
@@ -15,35 +15,33 @@
...
 
-Chaumian CoinJoin~\cite{mizrahi2013blind}\cite{maxwell2013coinjoin} is a technique to facilitate untrusted construction of collaborative Bitcoin transactions, also known as CoinJoins, by utilizing Chaumian blind signatures~\cite{chaum1983blind}. However this technique requires standard denominations, which limits how such transactions can be constructed.
+  This work introduces a new protocol for centrally coordinated CoinJoin implementations utilizing keyed verification anonymous credentials and homomorphic value commitments. This is an improvement on the state of the art in both privacy and flexibility, enabling some additional use cases and reduced overheads.
```
nopara73, line 25:
```suggestion
  This work introduces a new protocol for centrally coordinated CoinJoin implementations utilizing keyed verification anonymous credentials and homomorphic value commitments. This is an improvement on the state of the art in both privacy and flexibility, enabling additional use cases and reduced overheads.
```

main.tex
```diff
@@ -15,35 +15,33 @@
...
+\section{Introduction}
 
-\end{abstract}
+Bitcoin transactions redistribute funds by consuming as inputs the outputs of previous transactions, and creating new outputs. The protocol rules enforced by the network ensure that transactions do not arbitrarily inflate the monetary supply and that output are spent at most once. This can be simply verified because the amounts as well as the specific outputs being spent are broadcast as part of the transaction in the clear. This presents significant challenges for transacting with privacy\footnote{In this work we restrict the discussion of Bitcoin privacy to that of transactions, but there are other considerations especially at the network layer.}. % TODO cite Ron, Fistful of Bitcoins, Moser, ...?
```
nopara73, line 30:
```suggestion
Bitcoin transactions redistribute funds by consuming outputs of previous transactions as inputs, and creating new outputs. The protocol rules enforced by the network ensure that transactions do not arbitrarily inflate the monetary supply and that outputs are spent at most once. This can be verified because the amounts as well as the specific outputs being spent are broadcast as part of the transaction in the clear. This presents significant challenges for transacting with privacy\footnote{In this work we restrict the discussion of Bitcoin privacy to that of transactions, but there are other considerations especially at the network layer.}. % TODO cite Ron, Fistful of Bitcoins, Moser, ...?
```

main.tex
```diff
@@ -15,35 +15,33 @@
...
 
-% Important remainder from removed Motivation: In the current protocol input registration links the user's inputs, as well as their change output.
+The conditions for spending an output are specified in its \texttt{scriptPubKey}, typically requiring that the spending transaction be signed by a specific public key. The signatures authorizing a transaction usually commit to the transaction in its entirety, which makes it possible for mutually distrusting parties to jointly create transactions without any risk of funds being misallocated: participants will only sign a proposed transaction after confirming that their desired outputs are included.
```
nopara73, line 32:
```suggestion
The conditions for spending an output are specified in its \texttt{scriptPubKey}, typically requiring that the spending transaction to be signed by a specific public key. The signatures authorizing a transaction usually commit to the transaction in its entirety, which makes it possible for mutually distrusting parties to jointly create transactions without risk of funds being misallocated: participants will only sign a proposed transaction after confirming that their desired outputs are included.
```

main.tex
nothingmuch: stylistically this is out of place if we're going for an academic tone, if we want to claim that we must provide evidence, which seems out of scope, but with "may" there is no such burden

main.tex
nothingmuch: grammatically i'm not sure about this change, i think it needs a determinant of some type, but is "any" arguably too broad because risk is not eliminated entirely,... i think this needs to be rephrased a bit more

main.tex
nothingmuch: "those" is not grammatical in this context, what's the problem with that? also why remove common? i think it's relevant that they have been circulating widely and are commonly held

main.tex
nothingmuch: "simply" was added to contrast with e.g. cryptocurrencies where the verification is not simple, since strictly speaking it could be verified otherwise, i will rephrase so that it doesn't look like a fluff word

main.tex
nopara73: > "those" is not grammatical in this context

ACK.

> also why remove "common"

How about removing the "some" instead?

main.tex
nopara73: Isn't this an obviously true premise? If something has serious limitations on how it can be used, then it will be used less than without those limitations.

main.tex
nothingmuch: what do you think of this?

>   Bitcoin's transaction model is a publicly verifiable ledger of transferring value between pseudonymous identities. Although this design lacks strong guarantees for privacy, it does not rule it out.

shorter and more precise (doesn't make overly general statements), without actually adding unecessary details into the abstract

main.tex
nopara73: [Don't use simply.](https://jameshfisher.com/2017/02/22/dont-use-simply/)

> Uses of “simply” are an attempt to glorify the writer instead of comfort the reader. The subtext is “I the writer am smart. I find this simple. You should too. If you don’t, you’re stupid.”

main.tex
nopara73: My point was the word "pseudonymity" is red herring and we shouldn't use that to make blanket statements about Bitcoin. In fact it was a red herring from the very beginning: https://satoshi.nakamotoinstitute.org/emails/cryptography/threads/1/

> (Satoshi Nakamoto) It's not pseudonymous in the sense of nyms identifying people, but it is at least a little pseudonymous in that the next action on a coin can be identified as being from the owner of that coin.

main.tex
nothingmuch: it's just a stylistic convention that says the same thing, but qualifies that we are not going to try and argue for it empirically because that's out of scope. for example i'd expect that sentence without "may" in a usability study of bitcoin privacy software that shows a clear link with evidence. here it seems out of place for me, because it kind of blurs the boundaries between the definite claims we're making, assumptions we *need* to make to even state these claims (e.g. how bitcoin works), and non-critical assumptions, which this one is. this is a perfectly valid assumption to make, but it's not an assumption we rely on so adding "may" makes it clear that the validity of the claims we do make doesn't *depend* on it being true

main.tex
nothingmuch: > While some newer cryptocurrencies use more sophisticated approaches to defining such rules, in Bitcoin the amounts as well as the specific outputs being spent are broadcast in the clear as part of the transaction. This presents significant challenges for transacting with privacy

how's this?

nothingmuch: also for the record, while i agree with that post in general, this was not the same usage of "simply", it has nothing to do with the reader, only with the fact that in bitcoin these rules are simple because they don't use fancy crypto

main.tex
nothingmuch: > Bitcoin's model for value transfer is publicly verifiable ledger of transactions, with ownership of coins defined in terms of public keys.

how's this?

main.tex
MaxHillebrand:
```suggestion
  A number of methods have been proposed to mitigate these issues. Among these, CoinJoin is an approach to structuring transactions which adds ambiguity and breaks common assumptions that underlie heuristics used for surveillance.
```

PR #33, opened 2020-05-25
Add accent

nopara73:
Either all of us have accent or none of us.

### DIFF
--- a/main.tex
+++ b/main.tex
@@ -11,7 +11,7 @@
 \newtheorem{definition}{Definition}[section]
 
 \title{WabiSabi - draft v0.3}
-\author{Adam Ficsor, Yuval Kogman, István András Seres}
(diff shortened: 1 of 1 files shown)


nothingmuch:
can i have a heavy metal umlaut in that case? i need to decide where to put it

MaxHillebrand:
is it `Adam` now again? it should be have the accent...

nothingmuch:
oof, good catch! this conflicted with the s/draft/Draft/ change and i only noticed the accent on the o when i resolved the conflict.

nothingmuch:
@nopara73 since the bad fix is already, and it's really not any reviewer's business to decide how you want to spell your name, do you mind if you or i just fix directly in master?

nopara73:
Sure. Add your umlaut, too.

PR #39, opened 2020-05-26
Add limitations of ZeroLink to introduction

nothingmuch:
Introduces the main motivations for the design.

Includes some plots with data from https://github.com/nopara73/Dumplings

Co-authored-by: nopara73 <adam.ficsor73@gmail.com>
Co-authored-by: Seres István András <seresistvanandras@gmail.com>


nothingmuch:
Hmm, I think this is missing a definition of what "other" and "samourai" mean in the plots.

We should probably say something about what these groups of transactions are, and how they were identified, not just refer to the dumplings repo?

main.tex
```diff
@@ -39,7 +54,58 @@ \section{Introduction}
...
-\section{Our Contribution}
+\subsection{Limitations of ZeroLink}
+In this work, we are aiming to improve ZeroLink, the most popular Chaumian CoinJoin implementation for Bitcoin. We identify several privacy shortcomings and inefficiencies of the current protocol as implemented by Wasabi.
+
+\subsubsection{Round denominations}
+
+
+Due to the nature of blind signatures, mixed outputs of ZeroLink CoinJoins are restricted to a fixed set of value denominations which are multiples of a base denomination\footnote{approximately $0.1$\bitcoinA{}}.
+
+This creates friction when sending or receiving arbitrary amounts of Bitcoin, as the fixed denomination generally creates change which is smaller, both when mixing and when spending mixed outputs.
```
MaxHillebrand, line 65: I don't understand what you want to say here... that when you have a anonset coin worth 0.1, and you make a payment with it, the change is below 0.1 and thus you cannot mix again without consolidation?

main.tex
```diff
@@ -39,7 +54,58 @@ \section{Introduction}
...
+
+We define \emph{CoinJoin inefficiency} as the fraction of non-mixed outputs in a CoinJoin transaction, see Figure~\ref{fig:cjinefficiency}.
```
MaxHillebrand, line 67: if a change gets remixed, it's not inefficient. 

```suggestion
We define \emph{CoinJoin inefficiency} as the fraction of non-remixed change outputs in a CoinJoin transaction, see Figure~\ref{fig:cjinefficiency}.
```

main.tex
```diff
@@ -39,7 +54,58 @@ \section{Introduction}
...
+\begin{figure}[h!]
+    \centering
+    \includegraphics[scale=0.4]{Figures/CJInefficiency.pdf}
+    \caption{CoinJoin inefficiency of various privacy-focused Bitcoin wallets}
+    \label{fig:cjinefficiency}
+\end{figure}
+
+\subsubsection{Minimum denomination}
+
+In order to pariticpate a user's combined input amount are must be greater or equal to the base denomination. We observe, that considerable portion of CoinJoin inputs are less than this minimum denomination, see Figure~\ref{fig:minimumdenomination}.
```
MaxHillebrand, line 78:
```suggestion
In order to pariticpate a user's combined input amount must be greater or equal to the base denomination. We observe, that considerable portion of CoinJoin inputs are less than this minimum denomination, see Figure~\ref{fig:minimumdenomination}.
```

main.tex
```diff
@@ -39,7 +54,58 @@ \section{Introduction}
...
+\begin{figure}[h!]
+    \centering
+    \includegraphics[scale=0.4]{Figures/SmallValueInputsWasabi.pdf}
+    \caption{Fraction of input UTXOs with value smaller than the minimum denomination in Wasabi CoinJoin transactions}
+    \label{fig:minimumdenomination}
+\end{figure}
+
+Even if users are able to provide several smaller value input UTXOs with total value greater than the minimum denomination, the coordinator knows those input UTXOs belong to the same user.  In an ideal mixing protocol the coordinator should not obtain more information than the already available blockchain data by coordinating the CoinJoin transaction.
+
+Furthermore if users consolidate their UTXOs before the CoinJoin in an additional transaction in order to be able to participate in a CoinJoin, then this link is revealed publicly based on the common ownership heuristic~\cite{meiklejohn2013fistful}.
```
MaxHillebrand, line 89:
```suggestion
Furthermore if users consolidate their UTXOs before the CoinJoin in an additional transaction in order to be able to participate in a CoinJoin, then this link is revealed publicly based on the common input ownership heuristic~\cite{meiklejohn2013fistful}.
```

main.tex
```diff
@@ -39,7 +54,58 @@ \section{Introduction}
...
+
+\subsubsection{Variable denominations} Since users pay mining and coordination fees the denominations are gradually reduced between rounds in order to make it possible for users to mix several times without providing additional inputs. This introduces a perverse incentive to minimize coordination fees by remixing in quick succession in order, resulting in a smaller anonymity set % TODO how to say - because users are more "clumped together" in time
```
MaxHillebrand, line 91:
```suggestion
\subsubsection{Variable denominations} Since users pay mining and coordination fees the denominations are gradually reduced between rounds in order to make it possible for users to mix several times without providing additional inputs. This introduces a perverse incentive to minimize coordination fees by remixing in quick succession in order, resulting in a smaller anonymity set compared to time-staggered remixes.
```

main.tex
nothingmuch: yes, and also that when you have an arbitrary unmixed amount, in general you will get change which also must be consolidated if mixing

MaxHillebrand:
```suggestion
This creates friction when sending or receiving arbitrary amounts of Bitcoin, as the fixed denomination generally creates change which is smaller, both when remixing and when remixing the change of a payment.
```

nothingmuch: that change seems more confusing to me, when you are mixing some arbitrary amount you often get change because of the base denomination, and separately after mixing if you spend mixed outputs you get change for the same reason. you have to remix that change but this jumps ahead to change from remixing that change

main.tex
```diff
@@ -39,7 +54,57 @@ \section{Introduction}
 
 The use of standard denominations in the resulting CoinJoin transaction obscures the relationship between individual inputs and outputs, making the origins of each output ambiguous. Unfortunately standard denominations also present limitations, such as requiring change outputs from these CoinJoin transactions, and limiting the usefulness of the privacy enhanced outputs for payments of arbitrary amounts.
 
-\section{Our Contribution}
+\subsection{Limitations of ZeroLink}
+In this work, we are aiming to improve ZeroLink, the most popular Chaumian CoinJoin implementation for Bitcoin. We identify several privacy shortcomings and inefficiencies of the current protocol as implemented by Wasabi. Some metrics comparing Wasabi, Samourai and other apparent CoinJoin transactions are provided. ``Other'' includes JoinMarket, but also has an inherent false positive error given these transactions are identified heuristically.
+
+\subsubsection{Round denominations}
+
+Due to the nature of blind signatures, mixed outputs of ZeroLink CoinJoins are restricted to a fixed set of value denominations which are multiples of a base denomination\footnote{approximately $0.1$\bitcoinA{}}.
```
lontivero, line 62: This doesn't follow. What is the nature of blind signatures that makes zerolink denominations fixed?

nothingmuch: because a blind signature by a key can only convey 1 bit of information from the signer to the verifier

nopara73: I don't think it matters, but here you go my nitpicking:

- ZeroLink only has a single denomination.
- Wasabi has multiple denomination, which are the multiples of each other, but they don't have to be.

nothingmuch: i think we should rename the whole section to Limitations of Wasabi, and just say that wasabi implements a version of ZeroLink, i think that would be more accurate, and clearer for readers who vaguely know have heard of wasabi but don't know the details

PR #41, opened 2020-06-02
Add Scheduling overview

nopara73:


main.tex
```diff
@@ -159,6 +159,17 @@ \subsection{Terminology and notation}
 
 Finally, $k$ is a protocol level constant, denoting the number of credentials used in input and output registration requests, and $v_{\mathit{max}} = 2^{51}-1$ constrains the amount value ranges.\footnote{$\log_2(2099999997690000) \approx 50.9$}
 
+\subsection{Scheduling}
+
+A CoinJoin round consists of an Input Registration, an Output Registration and a Transaction Signing phases. The goal of WabiSabi is to ensure the inputs of users those do not comply with the protocol are identified, thus these inputs can be excluded from the following rounds in order to ensure completion of the protocol. Discussion of various blame strategies, methods to avoid timing attacks and other scheduling related topics are not part of this overview, rather we concentrate on identifying non-compliant inputs in various phases of the protocol:
```
MaxHillebrand, line 164:
```suggestion
A CoinJoin round consists of an Input Registration, an Output Registration and a Transaction Signing phases. The goal of WabiSabi is to ensure the inputs of users who do not comply with the protocol are identified, thus these inputs can be excluded from the following rounds in order to ensure completion of the protocol. Discussion of various blame strategies, methods to avoid timing attacks and other scheduling related topics are not part of this overview. Rather we concentrate on identifying non-compliant inputs in various phases of the protocol:
```

main.tex
```diff
@@ -159,6 +159,17 @@ \subsection{Terminology and notation}
...
+
+\begin{enumerate}
+    \item While identifying non-compliant inputs during Input Registration phase is trivial, there is no reason for issuing penalties at this point.
+    \item Identifying non-compliant inputs during Output Registration phase is not possible, thus this phase always completes and progresses to the Signing Phase.
```
MaxHillebrand, line 168: not sure if `inputs` is correct here?

```suggestion
    \item Identifying non-compliant outputs during Output Registration phase is not possible, thus this phase always completes and progresses to the Signing Phase.
```

main.tex
```diff
@@ -159,6 +159,17 @@ \subsection{Terminology and notation}
...
+    \item During Signing Phase, inputs those do not sign are the non-compliant inputs and they shall be issued a penalties.
```
MaxHillebrand, line 169:
```suggestion
    \item During Signing Phase, inputs those do not sign are non-compliant inputs and they shall be issued a penalties.
```

main.tex
```diff
@@ -159,6 +159,17 @@ \subsection{Terminology and notation}
...
+\end{enumerate}
+
+The point of the cryptographic primitives in WabiSabi is to ensure it does not worth for any non-compliant input to sign the final CoinJoin transaction, otherwise non-compliant inputs could not be identified. We achieve this by ensuring no user can register outputs those sum is larger than its registered inputs' sum.
```
MaxHillebrand, line 172:
```suggestion
The point of the cryptographic primitives in WabiSabi is to ensure that it is not possible for any non-compliant input to sign the final CoinJoin transaction, otherwise non-compliant inputs could not be identified. We achieve this by ensuring no user can register outputs those sum is larger than its registered inputs' sum.
```

main.tex
nopara73: Inputs are correct.

main.tex
nopara73: "It does not worth" the precise one as it's possible, they can sign their inputs they'd just lose money, so it does not worth it.

main.tex
nopara73: I can't spot the diff, but I've read your suggestion and it is correct, so I ACK it.

nopara73:
@nothingmuch @MaxHillebrand I addressed your reviews.

main.tex
nothingmuch: oops, i thought Max's original remark only addressed the grammar.

semantically i agree that it's about incentives, but it should say something like:

"ensure that it is not worth it for the owner of any non-compliant input to sign"

i have some small fixes i've been gathering based on others' feedback, i can add that in another PR i will make today

PR #43, opened 2020-06-03
Prove balance in zero knowledge

### LINKED ISSUE #40: Clarifying the Balance Proof
Hey,

I read through d498f645a1611205feff541ec8f8d7410f4fc9e9 and I like the design. Great to see that KVACs + Pedersen Commitments of value and serial number attributes + rangeproofs in fact lead to a working and simple system. I'm looking forward to see this in practice - in particular because it's not only applicable to Chaumian Coinjoin, but can also be used to implement ecash.

- Section 4.3 (Over-spending prevention by balance proof) is a bit unclear to me. Shouldn't `pi^sum[1]` be the sum of differences between `M_vi` and `C_vi` `r`-values?
  Moreover, as specified it seems like the user would really send `sum(z_i)` to the coordinator instead of providing a zero knowledge proof the `prod(M_vi/C_vi)` relation. This would be a problem, because after seeing `sum_z`, the coordinator can for all subsets of `M_vi` he has previously seen, test whether `G_v^sum_z prod(M_vi) ?= prod(C_vi)` and therefore link `M_vi` issuances with the credential presentation.
- I didn't have a look at how Perfect Hiding would work
- nit: footnote 8 on "small values of k": where do the values come from?
- nit: 4.2.1. (Credential Presentation) mention that `t_i` is from credential

nothingmuch:
This change addresses two mistakes found by Jonas Nick:

The first is a privacy leak. After seeing the sums of the r-terms the
coordinator could attempt link credential presentation to issuance. This
can be done by computing the products of different subsets of credential
request attributes, and seeing if the resulting commitment can be opened
to the sum of the `r` terms and the registered amount.

The second mistake was in the formula for the balance proof, neglecting
a set of r-terms.

Closes #40

### DIFF
--- a/main.tex
+++ b/main.tex
@@ -356,25 +356,21 @@ using its secret key (independently of the user's derivation), and verifies $\pi
 
 \subsection{Over-spending prevention by balance proof}\label{balance}
 
-The user needs to convince the coordinator that the amounts redeemed and the amounts requested differ by $\Delta_{v}$, which she can prove by including the following witness-hiding proof:
-\[ \pi^{\mathit{sum}}=\left( \sum_{i=1}^{k} r_{v_i}, \sum_{i=1}^k z_i\right) \]
+The user needs to convince the coordinator that the amounts redeemed and the amounts requested differ by $\Delta_{v}$, which she can prove by including the following proof of knowledge:
+\[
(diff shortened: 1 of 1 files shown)


MaxHillebrand:
Thanks @jonasnick for the report, and thanks @nothingmuch for the fix! :green_heart:

nothingmuch:
image

PR #48, opened 2020-06-07
[WIP] attacks section

nothingmuch:
this is still pretty rough, it's proven very hard to make this
discussion flow nicely without going into too much detail, and there are
still a few things that seem notable enough but that i can't find a good
place for.


nopara73:
This is not a good section IMO. You can barely say anything about attacks without the rest of the protocol.

nothingmuch:
yeah i've been struggling to make it work because of that. i guess we can either cut this from the draft, maybe coming back to it later,or try to fix it. i would prefer if the draft did at least mention some of these issues but what i have is too long, too vague, and i couldn't really find a way to organize it logically after moving things around countless times

nothingmuch:
updated version in #71

PR #49, opened 2020-06-07
Optimizations

### LINKED ISSUE #42: Eliminate serial number attribute
@RubenSomsen suggested a way to prevent double spending without needing a serial number attribute.

First, observe that instead of revealing the serial number `s`, with a new generator `G_r` knowledge of `s` can be proven revealing only a curve point `C_r = G_r^s`. This commitment does not contain `z`, so it's fixed for a single attribute and can be used as a serial number/nullifier if the proof also relates the witness `s` to `C_s`.

Next, to eliminate the serial number attribute we can use the `r` term from the amount commitment (`M_v = G_h^v G_g^r`) instead of `s`, to create `C_r = G_r^r`.

A proof of knowledge of `r` with a statement `(C_r, C_v / G_v^z G_h^v) = (G_r^r, G_g^r)` (DLEQ) is sufficient.

A `G_r` is required because if `G_g` is used the coordinator could brute force the small message space of values to deanonymize the scheme as described in #40.

Like serial numbers this does not provide unconditional hiding but that can be addressed if `M_v` includes additional randomness term with yet another distinct generator.

nothingmuch:
This change describes the the `U=HashToG(t)` optimization suggested by Greg Zaverucha and the optimization that eliminates the serial number attribute suggested by @RubenSomsen

Closes #42

### DIFF
--- a/main.tex
+++ b/main.tex
@@ -157,7 +157,7 @@ In order to facilitate construction of Bitcoin transactions, our credentials rep
 During credential presentation randomized versions of the attributes are presented, which we denote $C_a$ and $C_s$.
 \end{definition}
 
-Finally, $k$ denotes the number of credentials used in input and output registration requests, and $v_{\mathit{max}} = 2^{51}-1$ constrains the amount value ranges.\footnote{$\log_2(2099999997690000) \approx 50.9$}
+Finally, $k$ denotes the number of credentials used in input and output registration requests, and $a_{\mathit{max}} = 2^{51}-1$ constrains the amount value ranges.\footnote{$\log_2(2099999997690000) \approx 50.9$}
 
 \subsection{Phases}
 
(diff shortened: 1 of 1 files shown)


main.tex
```diff
@@ -14,6 +14,9 @@
 \usepackage[style=alphabetic,minalphanames=3,maxbibnames=99]{biblatex}
 \usepackage{hyperref}
 \usepackage{cleveref}
+\crefname{section}{\S}{\S\S}
+\Crefname{section}{\S}{\S\S}
```
nopara73, line 18: Not sure what this wants to be, but I have a hunch that you don't want to duplicate this line with different capitalization of "c"?

nothingmuch: there's separate commands for the middle and beginning of sentences, and in this case i'm overriding "section", "sections", "Section" and "Sections" to both be § and §§. \Cref is not used, so technically it's not needed but if we move things around this way it'll stay consistent

main.tex
```diff
@@ -466,79 +483,54 @@ \subsection{Credential Presentation}
...
 }
 \]
 using its secret key (independently of the user's derivation), and verifies $\pi_i^{\mathit{MAC}}$.
 
 \end{enumerate}
 
+\subsection{Double-spending prevention using serial numbers}\label{serial}
+
+The user proves that the group element $S_i$, which is used as a serial number, was generated correctly with respect to $C_{a_i}$:
+\[ \pi_{i}^{\mathit{serial}}=\operatorname{PK}\{ (z_ia_i,r_i): S_i = {G_s}^{r_i} \land C_{a_i} = {G_a}^{z_i}{G_h}^{r_i}{G_g}^{a_i} \} \]
```
lontivero, line 506:
```suggestion
\[ \pi_{i}^{\mathit{serial}}=\operatorname{PK}\{ (z_i,a_i,r_i): S_i = {G_s}^{r_i} \land C_{a_i} = {G_a}^{z_i}{G_h}^{r_i}{G_g}^{a_i} \} \]
```

main.tex
nopara73, line 18:
```suggestion
\Crefname{section}{\S}{\S\S}nothingmuch
```

nopara73: It was a test. Closing.

PR #53, opened 2020-06-09
[WIP] edits for clarity and flow

DanGould:


main.tex
```diff
@@ -39,30 +39,30 @@
...
 
 \subsection{Limitations of Wasabi}
-In this work, we are aiming to improve on ZeroLink~\cite{zerolink} as implemented by Wasabi, the most popular Chaumian CoinJoin implementation for Bitcoin. We identify several privacy shortcomings and inefficiencies of Wasabi CoinJoins. Some metrics comparing Wasabi, Samourai and other apparent CoinJoin transactions are provided. ``Other'' includes JoinMarket, but also has an inherent false positive error given these transactions are identified heuristically.
+In this work, we aim to improve on ZeroLink~\cite{zerolink} as implemented by Wasabi, the most popular Chaumian CoinJoin implementation for Bitcoin. We identify several privacy shortcomings and inefficiencies of Wasabi CoinJoins. Some metrics comparing Wasabi, Samourai and other apparent CoinJoin transactions are provided. The ``Other'' category includes JoinMarket, but also has an inherent false positive error given these transactions are identified heuristically.
 
-\subsubsection{Round denominations}
+\subsubsection{Fixed denominations}
```
nothingmuch, line 65: FWIW, I think this previously had "fixed" in the title, and I changed it because I felt like that implies it might be fixed between rounds. I don't feel very strongly either way but something to consider

main.tex
```diff
@@ -105,28 +105,28 @@ \subsubsection{Block-space efficiency}
...
 
 \section{Preliminaries}
 
 Hereby we give an informal and high-level description of applied cryptographic primitives. In the following the security parameter is denoted as $\lambda$.
 
 \subsection{Commitment schemes}
-A commitment scheme allows a party to commit to a message without enabling them to change their mind about the committed message after publishing the commitment. On the other hand the commitment should not reveal anything about the committed message.
+A commitment scheme allows one to commit to a chosen message while preventing them from changng the message after publishing the commitment. Thee commitment should not reveal anything about the chosen message.
```
nothingmuch, line 123:
```suggestion
A commitment scheme allows one to commit to a chosen message while preventing them from changng the message after publishing the commitment. These commitment should not reveal anything about the chosen message.
```

nothingmuch:
btw the commit message understates the contribution

main.tex
```diff
@@ -39,30 +39,30 @@
...
+  Although the public nature of this design lacks strong guarantees for privacy, it does not rule it out.
+  A number of methods have been proposed to create privacy. Among these, CoinJoin is an approach to structure transactions which adds ambiguity and breaks common assumptions that underlie heuristics used for surveillance.
+  Earlier implementations of CoinJoin transactions present a number of limitations which may partly explain the lack of their widespread adoption.
 
-  This work introduces a new protocol for centrally coordinated CoinJoin implementations utilizing keyed verification anonymous credentials and homomorphic value commitments. This is an improvement on the current approach which utilizes blind signatures in both privacy and flexibility, enabling additional use cases and reduced overheads.
+  This work introduces \emph{WabiSabi}, a new protocol for centrally coordinated CoinJoin implementations utilizing keyed verification anonymous credentials and homomorphic value commitments. This improvemes earlier approaches which utilize blind signatures in both privacy and flexibility, enabling novel use cases and reduced overhead.
```
yahiheb, line 49:
```suggestion
  This work introduces \emph{WabiSabi}, a new protocol for centrally coordinated CoinJoin implementations utilizing keyed verification anonymous credentials and homomorphic value commitments. This improves earlier approaches which utilize blind signatures in both privacy and flexibility, enabling novel use cases and reduced overhead.
```

main.tex
yahiheb: This should be: `These commitments` or `The commitment`.

main.tex
DanGould: Hmmm I read  "round denominations" to mean Round numbers https://en.bitcoin.it/wiki/Privacy#Round_numbers but I see what you mean as fixed between rounds. Maybe uniform denomination? I chose fixed cause that's what was described in the paragraph it titles

DanGould:
some questions/comments with approximate line numbers:

50. who does WabiSabi reduce overhead for? Non-obvious when it's brought up
65. what are we getting after here? that everyone has the same denomination right? not round numbers. "uniform CoinJoin denomination" or something similar might do it
91. I changed "blockchain" to "public ledger" to be consistent with our earlier language
114. why do we Capitalize Keyed-Verification Anon Cred here but not earlier? we should be consistent. Which one is it?

nothingmuch:
good questions
 
>     who does WabiSabi reduce overhead for? Non-obvious when it's brought up

the only benefit that can be argued directly from what is in the text is that the requirement to express your balance using the denominations is possibly removed. exactly how to do that is still out of scope for this iteration but i've been thinking about it, so maybe this is jumping the gun and we should just avoid mentioning anything...
 
>     what are we getting after here? that everyone has the same denomination right? not round numbers. "uniform CoinJoin denomination" or something similar might do it

Yes. The best I can come up with as a heading is "Denomination requirement", implying that some denomination is needed for the protocol without using the ambiguous word "round", but it doesn't strongly imply all participants share this denomination. "uniform", "fixed" or "constant" and to a lesser extent "standard" may imply that it's static between rounds until "variable" shows up later.

>     why do we Capitalize Keyed-Verification Anon Cred here but not earlier? we should be consistent. Which one is it?

hmm, not sure. i take responsibility and it is a matter of public record that i have no idea how to capitalize ;-)

somebody who is qualified should just decide and i'm happy to make sure it's consistent

nothingmuch:
just to be clear i don't feel strongly about that heading, "fixed denominations" is fine as far as i'm concerned

main.tex
```diff
@@ -105,28 +105,28 @@ \subsubsection{Block-space efficiency}
...
+WabiSabi can be instantiated to construct a variety of CoinJoin transaction structures that depart from the standard output denomination convention, used by SharedCoin\footnote{\url{https://github.com/sharedcoin/Sharedcoin}} and CashFusion\footnote{\url{https://github.com/cashshuffle/spec}} style transactions and Knapsack~\cite{maurer2017anonymous} mixing. Additionally it makes arbitrary consolidation of outputs possible, minimizes unmixed change, relaxes minimum required denominations, improves block space efficiency, and makes payments from as well as within CoinJoins, also known as PayJoins, possible.
```
nothingmuch, line 116: i think this is a bit less clear than the previous version, and this is exacerbated by my loose use of "payjoin" in the previous version.

i think a precise use is:

coinjoin = multiparty txn with the anonymity property
equal amount coinjoin = some subset of outputs having the same amounts
payjoin = steganographic coinjoin between 2 parties using arbitrary amounts and a net value transfer between parties

so what this makes possible in general is coinjoins, with few restrictions. we can require or allow equal amount coinjoins. if arbitrary amounts are allowed than arbitrary payments like joinmarket's send scripts can be done. better yet, value can be exchanged through credentials, allowing privacy enhanced net value transfers within a coinjoin, that have similarities to payjoins but which are not steganographic. the reason this is notable is that the ~0 balance assumption of the non-derived sub-transaction becomes much stronger with that as a possibility.

main.tex
```diff
@@ -164,15 +164,15 @@ \subsection{Credentials}
...
+To aid intuition we first describe a pair of protocols. One where credentials are issued during input registration, and then one where they're presented at output registration. $k$ denotes the number of credentials used in registration requests, and $a_{\mathit{max}} = 2^{51}-1$ constrains the range of amount values \footnote{$\log_2(2099999997690000) \approx 50.9$}. To improve privacy and efficiency these two cases are then generalized into a unified protocol used for both input and output registration, where every registration involves both presentation and issuance of credentials. This protocol is described in detail in \cref{details}.
```
nothingmuch, line 171: not sure if latex ignores this space or not tbh

```suggestion
To aid intuition we first describe a pair of protocols. One where credentials are issued during input registration, and then one where they're presented at output registration. $k$ denotes the number of credentials used in registration requests, and $a_{\mathit{max}} = 2^{51}-1$ constrains the range of amount values\footnote{$\log_2(2099999997690000) \approx 50.9$}. To improve privacy and efficiency these two cases are then generalized into a unified protocol used for both input and output registration, where every registration involves both presentation and issuance of credentials. This protocol is described in detail in \cref{details}.
```

main.tex
```diff
@@ -164,15 +164,15 @@ \subsection{Credentials}
...
 
-In order to facilitate construction of a CoinJoin transaction while protecting the privacy of participants we instantiate the scheme with a single group attribute $M_a$ which encodes a confidential Bitcoin amount as a Pedersen commitment. These commitments are never opened, instead properties of the underlying values are proven in zero knowledge, allowing the coordinator to verify validity of the requests made by honest participants. Ideally the coordinator would not learn anything that can't be learned from the resulting CoinJoin transaction.
+In order to facilitate construction of a CoinJoin transaction while protecting the privacy of participants, we instantiate the scheme with a single group attribute $M_a$ which encodes a confidential Bitcoin amount as a Pedersen commitment. These commitments are never opened. Instead, properties of the values they commit to are proven in zero knowledge, allowing the coordinator to validate the requests made by honest participants. Ideally the coordinator does not learn anything that can't be learned from the resulting CoinJoin transaction.
```
nothingmuch, line 167: "ideally the coordinator does not learn" seems a bit strong to me in the present tense, we can't actually guarantee that, especially when considering denial of service or other failures when a coinjoin tx isn't even produced, or things like de-registration, which reveals information to a coordinator about an input that doesn't end up participating

main.tex
```diff
@@ -88,11 +88,11 @@ \subsubsection{Minimum denomination}
...
+Even when users are able to provide several smaller value inputs with total value greater than the minimum denomination, the coordinator knows those inputs belong to the same user. In an ideal mixing protocol the coordinator should not obtain more information than the already available public ledger data by coordinating the CoinJoin transaction. This is especially important in practice as identifying sub-transactions when many multiple inputs and outputs are provided by the same parties is often computationally expensive using currently known algorithms, even without equal amounts being utilized~\cite{maurer2017anonymous}.
```
nothingmuch, line 91: this is not related to the change, but i thought i might open it up for discussion anyway

i think the computational complexity thing is a red herring, same as my criticism of cashfusion and same as the sharedcoin weaknesses: solving special cases of these problems is not so hard, especially when you care about constrained sub-txns. secondly, the attack gets exponentially easier as individual targets are de-anonymized. i don't think that's a valid hardness assumption and therefore it's not related.

what the knapsack paper does is more subtle, by construction it guarantees a minimum number of valid solutions that can be trivially found. instead of making it hard for the adversary to find one solution, finding solutions is easy, the problem is deciding between them without relying on external information

main.tex
DanGould: you're right. I misread this section when editing and the changing such as to using gets especially unclear

main.tex
DanGould: What about "Ideally the coordinator ... CoinJoin Transaction. But in practice when considering DoS or other failure that might cause a CoinJoin not to be produced some information about an input that does not participate may be shared" ... and explain how that effects privacy? Maybe too long winded

main.tex
DanGould: what is a sub-transaction in this line? does it mean one of multiple inputs and outputs provided by the same party? or one of the transactions of the protocol as a whole. Sorry, I got lost here

main.tex
nothingmuch: see #48 but that's probably not going to make it into this as per discussion...

it's not wrong, ideally that would be the case, but it'd be better if it didn't imply that this was accomplished

main.tex
nothingmuch: the first one, defined in the knapsack paper (`maurer2017anonymous` citation)

main.tex
```diff
@@ -408,7 +408,7 @@ \section{Cryptographic Details}\label{details}
 
 \subsection{Credential Requests}
 
-For each $i \in [1, k]$ the user chooses an amount $0 \leq a_i < a_{\mathit{max}}$ subject to the constraints of the balance proof (\cref{balance}).
+For each $i \in [1, k]$ the user chooses an amount $a_i \mid 0 \leq a_i < a_{\mathit{max}}$ subject to the constraints of the balance proof (\cref{balance}).
 
 She commits to the amount with randomness $r_i \in_R \mathbb{Z}_q$, and this commitments is the attribute of the requested credentials.:
```
nothingmuch, line 413:
```suggestion
She commits to the amount with randomness $r_i \in_R \mathbb{Z}_q$, and these commitments are the attributes of the requested credentials.:
```

main.tex
nothingmuch:
```suggestion
Even when users are able to provide several smaller value inputs with total value greater than the minimum denomination, the coordinator knows those inputs belong to the same user. In an ideal mixing protocol the coordinator should not obtain more information than the already available public ledger data by coordinating the CoinJoin transaction. This information removes many degrees of freedom when assigning non-derived sub-transactions~\cite{maurer2017anonymous}, potentially removing ambiguity when there are multiple valid assignments or reducing the computational cost of such an analysis.
```

main.tex
nothingmuch:
```suggestion
WabiSabi can be instantiated to construct a variety of CoinJoin transaction structures that depart from the standard output denomination convention, used by SharedCoin\footnote{\url{https://github.com/sharedcoin/Sharedcoin}} and CashFusion\footnote{\url{https://github.com/cashshuffle/spec}} style transactions and Knapsack~\cite{maurer2017anonymous} mixing. Payments from CoinJoin transactions are possible, as are payments within them making this potentially a multiparty PayJoin that trades the steganographic properties for improved privacy from counterparties. Additionally, restrictions on consolidation of inputs can be removed, and there are opportunities for reducing unmixed change and relaxing minimum required denominations, and improved block space efficiency.
```

main.tex
```diff
@@ -164,15 +164,15 @@ \subsection{Credentials}
...
 
 In order to maintain privacy clients must isolate registration requests using unique network identities. A single network identity must not expose more than one input or output, or more than one set of requested or presented credentials.
 
-For fault tolerance request handling should be idempotent, allowing requests whose response was lost to be retried without modification, using the same or a new network identity.
+For fault tolerance, request handling should be idempotent, allowing requests whose response was lost to retry without modification, regardless of network identity.
```
nothingmuch, line 175: not regardless! either the old network identity can be retried, or a new one must created, but if any other existing network identity that will link information about linkage. also the requests themselves don't retry, they are retried by the client

```suggestion
For fault tolerance, request handling should be idempotent, allowing a client to retry failed requests without modification, using the same network identity as previous attempts or a fresh one of network identity.
```

main.tex
nothingmuch: when in doubt, delete

```suggestion
\subsubsection{Denominations}
```

nopara73: Fixed denomination is incorrect. Not only it is incorrect inter-rounds, but also intra-rounds as the announced denomination almost always differs from the denomination that the final TX will eventually have.  
So fixed denomination is incorrect, and round denomination is misleading. Then let's delete.

main.tex
```diff
@@ -79,7 +79,7 @@ \subsubsection{Round denominations}
 
 \subsubsection{Minimum denomination}
 
-In order to pariticpate a user's combined input amount must be greater or equal to the base denomination.\footnote{The observed base denominations in Wasabi's CoinJoins is usually slightly higher than the announced, agreed upon base denomination. Thus participants sometimes get back slightly more value in the CoinJoins than how much they put into.} We observe, that considerable portion of CoinJoin inputs are less than this minimum denomination, see \cref{fig:minimumdenomination}.
+In order to pariticpate a user's combined input amount must be greater or equal to the fixed base denomination.\footnote{The observed base denominations in Wasabi's CoinJoins is usually slightly higher than the announced, agreed upon base denomination. Thus participants sometimes get back slightly more value in the CoinJoins than they put in.} We observe, that considerable portion of CoinJoin inputs amount to less than this minimum denomination, see \cref{fig:minimumdenomination}.
```
nopara73, line 82:
```suggestion
In order to pariticpate a user's combined input amount must be greater or equal to the base denomination.\footnote{The observed base denominations in Wasabi's CoinJoins is usually slightly higher than the announced, agreed upon base denomination. Thus participants sometimes get back slightly more value in the CoinJoins than they put in.} We observe, that considerable portion of CoinJoin inputs amount to less than this minimum denomination, see \cref{fig:minimumdenomination}.
```

main.tex
nopara73: Just a note: instead of sub-tx it's somewhat equivalent to say sub-mapping (and indeed Maurer et al say that, too.)

PR #60, opened 2020-06-11
Fix typos

yahiheb:
> The limit on participant count is because 274 and 124 are the minimum weight units

This sentence doesn't seem right.


main.tex
```diff
@@ -281,7 +281,7 @@ \subsection{Examples}
     \path[->] (i) edge node {\tiny{7}} (o1);
     \path[->] (i) edge node {\tiny{3}} (o2);
   \end{tikzpicture}
-  \caption{Alice wants to spend an input of amount 10 and create two outputs with amounts 7 and 3 (e.g. a payment and change)}
+  \caption{Alice wants to spend an input of amount 10 and create two outputs with amounts 7 and 3 (e.g. a payment and a change)}
```
nothingmuch, line 284: reverting this because "change" is a mass noun

PR #61, opened 2020-06-11
Last minute grammar fixes & rewording for public draft

nothingmuch:
Here are my last fixes, and the draft of the announcement email:

> As part of research into how CoinJoins in general and Wasabi in particular can be improved, we'd like to share our new building block WabiSabi, which utilizes keyed verification anonymous credentials instead of blind signatures to verify the honest participation of users in a centrally coordinated CoinJoin protocol with improved privacy and flexibility.
>
> Blind signatures have been used to facilitate centrally coordinated CoinJoins, but require standard denominations, each associated with a key, because blind signatures can only convey a single bit of information from the signer to the verifier (both roles are the coordinator in this setting). Anonymous credential carry attributes, and in our case these carry homomorphic value commitments as in Confidential Transactions.
>
> Note that this is an early version with a deliberately narrow scope, and only introduces this abstract building block.. At this stage we'd like to solicit feedback and criticism about our scheme and inputs with regards to its potential applications before proceeding. We do not not (yet) address the structure of the CoinJoin transactions, fee structures, or other implementation details, but discussion of these aspects is welcome.
>
> The repository is https://github.com/zkSNACKs/WabiSabi, and the latest version is available here: https://github.com/zkSNACKs/WabiSabi/releases/latest/download/WabiSabi.pdf


DanGould:
re: mail

> ...coordinator in this setting). Anonymous **credentials** carry "attributes", in our case, homomorphic value commitments as in Confidential Transactions.

>Note that this is an early version with a deliberately narrow scope, and only introduces this abstract building **block.**

?

DanGould:
or "Each Anonymous credential carries 'attributes'"

nothingmuch:
OK i'm just going to merge this, i fixed those issues in the email and got an ACK on slack, and since it's already late Max and Adam can complain tomorrow ;-)

PR #62, opened 2020-06-15
Create Scheduling.md

nopara73:
This is a an idea of how we could progress working out Scheduling. IMO a good way would be to describe a simple scheme that works, then keep bringing up issues and extending the scheme to resolve them. Thoughts?

(For those who cannot figure out how to view the md preview: https://github.com/zkSNACKs/WabiSabi/blob/nopara73-patch-1/Scheduling.md)

### DIFF
--- a/Scheduling.md
+++ b/Scheduling.md
@@ -1,0 +1,72 @@
+# Scheduling
+
+## Overview
+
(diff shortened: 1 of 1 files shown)


nopara73:
Moved this to an issue as we discussed: https://github.com/zkSNACKs/WabiSabi/issues/64

PR #68, opened 2020-06-26
Add StatusRequest and Coordinator parameters explanations

nopara73:


nopara73:
Sorry, but I don't understand.

nothingmuch:
the lines i link to talk about how the details of the status message are not yet decided, and discussion is ongoing in the github thread, but now you've added them... i was trying to think of a suggested edit but didn't come up with anything:

- the parenthesized link in the previous text i linked to can just be removed (KISS)
- the text you added could be moved up from the interaction diagram section to this section (i think this makes more sense, because the status request doesn't pertain to the diagram, it stands on its own)
- the previous text can just be removed altogether, because it doesn't really add anything and the document flow can be improved later
- the previous text can be rewritten so that everything is consistent (i tried and couldn't come up with an idea)

i couldn't decide which makes the most sense overall, but right now it's kind of redundant and internally inconsistent

nopara73:
> the lines i link to talk about 

Is broken. You're linking to L63-L65, but the PR starts at L107 thus your link brings up the whole diff between this PR and the master.

> talk about how the details of the status message are not yet decided

We can't decide before amount organization takes place, but...

> right now it's kind of redundant and internally inconsistent

...but has no chance to understand the rest of the diagrams without knowing what's going on with the status messages, thus it's not redundant. In fact to resolve the inconsistency we should remove the "the lines i link to talk about how the details of the status message are not yet decided". And this way it'll only be incomplete, but not inconsistent and when amounts come in we can extend and modify it accordingly.

nothingmuch:
yeah I'm talking about lines that should have been changed but weren't,
they talk about how what you added as if it's not yet added

nopara73:
> the text you added could be moved up from the interaction diagram section to this section

Not sure. It'd be more difficult to read without the relevant diagram IMO.

> the previous text can just be removed altogether, because it doesn't really add anything and the document flow can be improved later

What do you refer to as "previous text"?

nothingmuch:
> What do you refer to as "previous text"?

the text i linked to... i should have said "out of date text"

PR #69, opened 2020-06-26
Abstract away Tor

nopara73:


nothingmuch:
LGTM, but at some later point I think we do want reintroduce this specificity, since Tor is the only anonymity network that will be used in the foreseeable future, and it affects the threat model WRT global network level adversaries.

PR #71, opened 2020-07-04
CES updates

nothingmuch:
changes made for CES submission, can be broken down into separate changes for easier review


main.tex
```diff
@@ -521,11 +560,118 @@ \subsection{Over-spending prevention by balance proof}\label{balance}
...
+
+\subsection{Availability}
+
+\subsubsection{Malicious Coordinator}
+
+Being a central point of failure, the coordinator is a trusted party with regards to availability. If competing coordinators charge fees for their services then this is a minimal assumption in practice given the financial incentive.
+
+A malicious coordinator can fully disrupt the protocol by censoring certain inputs either at input registration or during the signing phase. Such denial of service can amplify attacks on privacy, by partitioning users in order to perform set intersection attacks. A malicious coordinator can drop messages causing any user to appear to be non-compliant, and therefore disrupt the protocol arbitrarily, and always learns the requested inputs and outputs, even if a round fails.
```
nopara73, line 581:
```suggestion
A malicious coordinator can fully disrupt the protocol by censoring certain inputs either at input registration or during the signing phase. Such denial of service can amplify attacks on privacy, by partitioning users in order to perform set intersection attacks. A malicious coordinator can drop messages causing any user to appear to be non-compliant, and therefore disrupt the protocol arbitrarily, and always learns the requested inputs and outputs, even if a round fails. The cost of this attack depends on the Bitcoin network fees and the number of inputs and outputs provided by the coordinator. For example if the network fees paid in the CoinJoin are 1 USD for every input and output and the number of inputs and outputs are 100-100, then the coordinator needs 2*99 USD to execute this attack for one UTXO.
```

nopara73: I'm not sure my comment is academic enough, but the point is it's far from costless as the current text suggests.

main.tex
```diff
@@ -521,11 +560,118 @@ \subsection{Over-spending prevention by balance proof}\label{balance}
...
+
+The goal of the protocol is to allow a coordinator to provide the service to honest participants, without learning anything about the mapping between registered input and outputs, apart from what is already deducible given the public amounts visible on the Bitcoin blockchain. WabiSabi leverages the unlinkability of anonymous credentials and the hiding property of the amount commitments to minimize privacy leaks when a set of participants utilizes a centralized coordinator to reach agreement about such a transaction.
+
+\subsection{Availability}
+
+\subsubsection{Malicious Coordinator}
+
+Being a central point of failure, the coordinator is a trusted party with regards to availability. If competing coordinators charge fees for their services then this is a minimal assumption in practice given the financial incentive.
```
nopara73, line 579: I think this is what you meant.

```suggestion
\subsection{Availability}

Being a central point of failure, the coordinator is a trusted party with regards to availability. If competing coordinators charge fees for their services then this is a minimal assumption in practice given the financial incentive.

\subsubsection{Malicious Coordinator}
```

nopara73: On another note, I think it's incorrect. I don't think financial incentive is the key here. I think the key is whatever incentive the coordinator has to provide this service is inherent in mitigating this availability attack.

It's a bit circular in the sense that the incentive to be online/be available is the same as the incentive to run the service. The incentive to keep running the service is the incentive to run the service. Not sure how to put this into words, but it's just something very inherent that whoever runs the coordinator will try to maximize uptime.

main.tex
```diff
@@ -521,11 +560,118 @@ \subsection{Over-spending prevention by balance proof}\label{balance}
...
+
+\subsubsection{Malicious Users}
+Signatures can only be made after a transaction has been negotiated, and all inputs must provide a valid signature. Consequentially users can always disrupt the protocol during the final phase. Failure to sign is attributable to specific inputs and therefore can be mitigated by the coordinator. This allows the remaining honest participants to restart the protocol. Denial of service is not costless because unspent transaction outputs are a limited resource.
+
+A malicious participant still learns all of the input and output registrations\footnote{If \texttt{SIGHASH\_ANYONECANPAY} is set in the signature flags the full set of inputs could be kept known only to the coordinator until all signatures have been provided.} including amounts and scripts, so denial of service also presents privacy concerns.
```
nopara73, line 586: Expand on sybil by malicious users to be complete. Make sure that 

- [ ] malicious coordinator availability attacks are discussed - you did this, see my comment above why it's not good
- [ ] malicious coordinator sybil is discussed - you did this, see my comment why I think it's misleading
- [ ] malicious peers availability attacks are discussed - you did this
- [ ] malicious peers sybil is discussed - you didn't do this (well, a few words on spying on incomplete transactions, which is ok, but not complete for full sybils)

Consequently you should fix up the section names.

main.tex
```diff
@@ -521,11 +560,118 @@ \subsection{Over-spending prevention by balance proof}\label{balance}
...
+
+The coordinator observes registration requests according to a partial order which is an extension of the partial order defined by the DAG, in other words the coordinator knows that parallel requests do not share an edge and that dependent requests must be made sequentially. Clients can mitigate these leaks by minimizing dependencies between requests, scheduling requests randomly during phases of a known duration, and making additional reissuance requests.
+
+\subsubsection{Active attacks}
+
+Sybil attacks~\cite{douceur2002sybil} are inherent to mixing schemes, because a transaction between $n$ apparent participants $n-1$ of which are controlled by an attacker will fully link the victim's coins on both sides of the CoinJoin while giving the impression that the victim's privacy has been improved. There is a liquidity requirement for such an attack since participants must provide valid inputs\footnote{See also JoinMarket fidelity bonds: \url{https://gist.github.com/chris-belcher/18ea0e6acdb885a2bfbdee43dcd6b5af}}, as well as a cost imposed by mining fees.
```
nopara73, line 602: Ok, now this is problematic regarding how this section is here and for some reason the coordinator sybiling isn't an active attack anymore? Can you sort this out based on my above comments before I go further with the review?

main.tex
```diff
@@ -521,11 +560,118 @@ \subsection{Over-spending prevention by balance proof}\label{balance}
...
+ 
+\begin{description}
+ \item[Denomination] Some of the proposed and deployed schemes only support fixed amounts. Support for variable amounts typically add complexity, reduces privacy, or both.
+ \item[Performance] Fully decentralized schemes, CoinJoin based or otherwise, typically have higher communication complexity and therefore support fewer participants compared to centralized ones. CoinJoin-based mixing protocols provide mixing in a single Bitcoin transaction. Non-CoinJoin schemes often require more block space, for instance Xim~\cite{bissias2014sybil} requires $7$ on-chain transactions, TumbleBit~\cite{heilman2017tumblebit} $4$ transactions, Blindcoin~\cite{valenta2015blindcoin} and Mixcoin~\cite{bonneau2014mixcoin} both require $2$ transactions. Long sequence of on-chain transactions result in longer delays for users.
+ \item[Infrastructure] Several mixing-protocols assume and rely on infrastructure which is not practically available. For example, CoinShuffle assumes a peer-to-peer public bulletin board.
```
nopara73, line 617:
```suggestion
 \item[Infrastructure] Several mixing-protocols assume and rely on infrastructure which is not practically available. For example, CoinShuffle and CoinShuffle++ assume peer-to-peer public bulletin boards.
```

Even though CS implementations are actually using coordinators like us instead of the bulletin board approach. (Factcheck me.)

main.tex
```diff
@@ -521,11 +560,118 @@ \subsection{Over-spending prevention by balance proof}\label{balance}
...
+	CoinShuffle++~\cite{ruffing2017p2p} & $\mathcal{O}(n)$ & fixed & decentralized & n.a. & \cmark\footnotemark[\value{footnote}] \\
+	CashFusion & $\mathcal{O}(n)$ & variable & centralized & \cmark\footnote{Assuming server does not participate as a verifier.} & \cmark\footnote{Transaction ambiguity guarantees are probablistic, with additional computational difficulty under assumptions.} \\
+	JoinMarket & $\mathcal{O}(n)$ & variable & by taker\footnote{Transactions are initiated and coordinated by users buying mixing offers on a decentralized markeplace.} & n.a.\footnotemark[\value{footnote}] & \cmark \\
+	Chaumian CoinJoin & $\mathcal{O}(n)$ & fixed & centralized & \cmark & \cmark \\
+	This Work & $\mathcal{O}(n)$ & variable & centralized & \cmark & \cmark \\
+\end{tabular}
+\end{minipage}
+\caption{CoinJoin based protocols, no theft risk and requiring a single overt transaction.}
+\label{table:coinjoin}
+\end{table}
```
nopara73, line 645: Can you put a screenshot here? I can't review it without looking at it :)

main.tex
```diff
@@ -521,11 +560,118 @@ \subsection{Over-spending prevention by balance proof}\label{balance}
...
+	CoinSwap\cite{maxwell2013coinswap} & $\mathcal{O}(1)$\footnote{$n=2$} & 4\footnotemark[\value{footnote}] & var. & \cmark & p2p & n.a. & \xmark & disjoint \\
+	Xim~\cite{bissias2014sybil} & n.a. & $7$\footnotemark[\value{footnote}] & var. & \cmark &  p2p\footnote{Advertisements are made on the blockchain} & n.a. & \xmark & disjoint \\
+	CoinParty~\cite{ziegeldorf2015coinparty}& ?$\mathcal{O}(n \log n)$? & 2n & fix. & \cmark\footnote{Assuming $\frac{1}{3}$ of participants are honest.} & dec. & n.a. & \cmark\footnote{Assuming at least 2 participants are honest.} & covert \\
+	CoinSwap\cite{belcher2020design} & $\mathcal{O}(n)$ & $\mathcal{O}(n)$ & var. & \cmark &  taker & n.a. & \cmark & covert \\
+	% A2L?
+\end{tabular}
+\end{minipage}
+\caption{Non-CoinJoin based protocols.}
+\label{table:noncoinjoin}
+\end{table}
```
nopara73, line 669: Same, please screenshot.

main.tex
```diff
@@ -521,11 +560,118 @@ \subsection{Over-spending prevention by balance proof}\label{balance}
...
+
+A malicious coordinator could also delay processing of requests in order to gain more through timing and ordering leaks. In the worst case the coordinator can attempt to linearize all requests by blocking and unblocking individual users to recover full set of labeled edges. This is possible when $k=1$ and users have minimal dependencies between their requests and tolerate arbitrary timeouts but issue requests in a timely manner.
+
+A malicious coordinator may also tag users by providing them with different issuer parameters. When registering inputs a proof of ownership must be provided. If signatures are used, by covering the issuer parameters and a unique round identifier these proofs allow other participants to verify that everyone was given the same parameters.
+
+\section{Related Work}\label{sec:relatedwork}
```
nopara73, line 610: Kinda wanna add a few words on ecash literature?

main.tex
```diff
@@ -39,28 +59,34 @@
...
+Existing implementations of CoinJoin present a number of limitations which may partly explain the lack of their widespread adoption.
 
-  This work introduces \emph{WabiSabi}, a new protocol for centrally coordinated CoinJoin implementations utilizing keyed verification anonymous credentials and homomorphic value commitments. This improves earlier approaches which utilize blind signatures in both privacy and flexibility, enabling novel use cases and reduced overhead.
+This work introduces \emph{WabiSabi}\footnote{The Japanese concept \emph{wabi-sabi} is an aesthetic world view which among other things emphasizes acceptance of imperfections, in our case Bitcoin's challenges for privacy, and an appreciation for simplicity and economy.}, a new protocol for centrally coordinated CoinJoin implementations utilizing keyed verification anonymous credentials and homomorphic value commitments. This improves earlier approaches which utilize blind signatures in both privacy and flexibility, enabling novel use cases and reduced overhead.
```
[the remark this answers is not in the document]
nothingmuch: why should that be removed?

main.tex
nothingmuch: a conflicting incentive might be to prevent people using these services entirely in order to keep bitcoin surveillable, drowning out legitimate coordinators with ones that just fail to work, and run at a loss, but that doesn't seem sustainable at scale

main.tex
nothingmuch: i think your remark applies to sybil attacks, but this only addresses denial of service (attacks on availability)

main.tex
nothingmuch: image

main.tex
nothingmuch: image

main.tex
nothingmuch: i was considering it but it's very broad and we don't really take any concepts from it that isn't already cited, maybe apart from Chaum's definition & Brands on anon credentials (like ACL paper's introduction)

main.tex
nothingmuch: note that CoinParty was omitted from the submitted version, and i still need to go over and verify these one, i suspect there are still some inaccuracies

nothingmuch: thinking about this table a bit more:

txn should be split into confirmations and something that scales with the number of coins, the number of txns doesn't matter per se for the footprint, and in some cases having `n` as a term is misleading (e.g. coinparty requires 2 confirmation phases, not `n`, and the footprint is on the same order as that of an equivalent coinjoin transaction)

nothingmuch: @seresistvanandras thoughts?

main.tex
nothingmuch: these are both sub-subsections of the "availability" subsection, i.e. denial of service attacks

sybil attacks are covered in the active attacks sub-subsection of the "unlinkability" subsection

main.tex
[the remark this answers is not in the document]
nopara73: Makes the story more consistent.

main.tex
nopara73: Aah, I missed that.

main.tex
nopara73: Got it.

main.tex
nopara73: ACK

main.tex
nopara73: I can't really verify this table is correct and I think ti'd be somewhat counterproductive spending my time on this if the alternative is to progress forward WabiSabi, so I'll ACK it based on I trust your judgements.

main.tex
nopara73: Fine. It'd be nice, but also somewhat of a waste of time.

main.tex
nopara73: I resolve my attacks section comments and review it again. I had some confusion, but now more things are clear.

main.tex
nopara73: I resolve my attacks section comments and review it again. I had some confusion, but now more things are clear.

main.tex
```diff
@@ -521,11 +560,118 @@ \subsection{Over-spending prevention by balance proof}\label{balance}
...
+
+\section{Security and Privacy} \label{sec:securitayandprivacy}
+
+In this section, we discuss the security and privacy guarantees of the WabiSabi credential scheme for construction of CoinJoin transactions. Theft concerns addressed through Bitcoin's security model, making WabiSabi trustless in that regard. Since CoinJoins are an overt technique privacy strongly depends on the structure of the transactions themselves. WabiSabi is designed as a general purpose mechanism, so those details are outside of the scope of this work.
+
+The goal of the protocol is to allow a coordinator to provide the service to honest participants, without learning anything about the mapping between registered input and outputs, apart from what is already deducible given the public amounts visible on the Bitcoin blockchain. WabiSabi leverages the unlinkability of anonymous credentials and the hiding property of the amount commitments to minimize privacy leaks when a set of participants utilizes a centralized coordinator to reach agreement about such a transaction.
+
+\subsection{Availability}
+
+\subsubsection{Malicious Coordinator}
```
nopara73, line 577: If you write subsections one after another that means the second subsection will be a subsection of the first subsection?

nothingmuch: one is a subsection, the other is a subsubsection

PR #73, opened 2020-07-07
Add deadlines

nopara73:


protocol.md
```diff
@@ -107,7 +107,7 @@ The Interaction Diagram denotes Satoshi actors those are polling the Coordinator
 
 - `Phase`
 - `RoundId`
-- `Timeout`
+- `PhaseDeadline`
```
nothingmuch, line 110: ISO 8601? unix epoch?

nopara73: Yup. I think that's a detail, but I found that to be the most convenient to use except when I want to look at it with my own eyes.

Issue #18, opened 2020-04-24
Pre-paying the coordinator fee


MaxHillebrand:
This might be too early to bring up, but I have a couple ideas on the coordinator fee, and am curious if the crypto magic of the current protocol would hold up for it. 

## tl;dr

unlinkable pre-payment of the coordinator fee across rounds.

### avoid creating small value coins

In a self-spend, say the user has inputs worth `1.01 btc`, and the fee is `0.005 btc`, then the user would receive `1 btc` anonset output, no on-chain change, but instead a fee pre-payment worth `0.005 btc`, which can be used for the next round.

I have an intuition that this would make subset sum analysis on non-equal-value outputs even more difficult.

### getting rid of small value coins

If a user has a small value coin, say `0.01 btc`, then for most real payments he would have to consolidate it with another coin, thus revealing common ownership. Instead, he can register this small coin in the input of a CoinJoin, without receiving any output, the value goes directly to the coordinator fee output. In exchange he gets `0.01 btc` fee pre-payment that can be used in a subsequent round to pay for the fee.

### gift CoinJoin to others

The coordinator could gift fee pre-payment to users, as a reward for contributions, or as marketing, or for whatever reason. If the pre-payment can be transferred between users, then this could be a further opportunity to gift rounds to others.

## Privacy benefits

I think this would mess with assumptions of the CoinJoin quite a bit. The details of course depend on how ever we are going to structure the tx. But in general...

- some users have inputs, but no outputs
- some users have value inputs = outputs
- some users have value inputs < outputs

## Legal aspect

I spoke with Balint, and in general, having a fee pre-paid giftcard is not of legal concern, this is not a custodianship. Rather, it is debt on the companies balance sheet, denominated in bitcoin, and it is settled in the providing of the coordination service. There is no payout of pre-paid giftcards to bitcoin directly. [that would be the users has no inputs, but has outputs, and provides the giftcards off-chain.]

## Crypto magic

Here I need Your help with clarification if something like this is possible. The idea is that the coordinator can give out a token with a certain value. This token can be used in input registration to prove that the fee is paid. If possible, without the coordinator finding out that such a gift card was used. Or at least, that he does not know which gift card was used.

nothingmuch:
[...]

PR #74, opened 2020-07-08
Discuss fee options in protocol document

nothingmuch:
- Client calculation network & coordinator fees and account in balance proof
- Coordinator adjustment of credential amounts

### DIFF
--- a/protocol.md
+++ b/protocol.md
@@ -95,6 +95,22 @@ users or the coordinator have a liquidity requirement cost requirements dependin
 Mitigation of such attacks are beyond the scope of this protocol, as they
 pertain to the specific transaction structure.
 
+## Fees
+
+Mining fees per registration can be specified as a linear function of the
+weight units of the registration with specified precision and rounding
+behaviour. If there is no constant term (fixed per input/output fee), then this
+is just a conversion factor from weight units to satoshis, otherwise a pair of
+numbers. Coordination fees can also be specified as a linear function of the
+amount but may require a more detailed function. In either case the balance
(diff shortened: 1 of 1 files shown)


nothingmuch:
after transaction structure we will probably have to revisit this if the coordinator provides incentives for certain outputs. for example, a policy i have in mind is to allow 0 mixing fees for sufficiently old, low hamming weight outputs of distinct previous coinjoins by the same coordinator, effectively letting users remix for free but only if they've aged their mixed outputs enough. it's not clear to me if it's better to make that be a static policy or done by adjustments. also note that the coordinator can pre-announce any adjustment at input registration instead of waiting until the last connection confirmation to reveal the adjustment

nothingmuch:
note that this does not cover #18, which would require a more complex design:

- long lived credential fees (can just be wabisabi credentials with per epoch issuer parameters)
- additional presentation and change credential request with these fee tokens
- instead of one balance term, there are two, one for the nominal amount and one for fees paid (both network and coordinator fees)
- input amounts can be converted to fee amounts but not vice versa
- fee credentials are not actually issued until after the round fails or the coinjoin tx is confirmed. after failure, presented credentials can be refunded in exchange for new ones. on success change requests previously validated by balance proofs can be issued.
- fee reissuance command for consolidation and epoch migration (older tokens expire to keep server state bounded)
- the coordinator would cover mining fees by providing inputs, which it can do as part of consolidation of previous transactions' fees

such a design would allow small amounts of change to be converted to prepaid fee tokens

PR #75, opened 2020-07-10
minor fixes

MaxHillebrand:
just a couple nits I found while re-reading.


MaxHillebrand:
> Since CoinJoins other than PayJoin

I think as of now, the only steganographic CJ is PayJoin. But hopefully more will be found and implemented in the future. So this isn't good.

> Since equal value CoinJoins

Is not accurate, as even unequal amount cj can be fingerprinted, like for example knapsack. So this isn't good.

> Since most CoinJoins

That's why I think that this is an accurate wording. But probably there's a better prose in another way too.

main.tex
```diff
@@ -581,7 +581,7 @@ \subsubsection{Malicious Coordinator}
...
 
 \subsubsection{Malicious Users}
-Signatures can only be made after a transaction has been negotiated, and all inputs must provide a valid signature. Consequentially users can always disrupt the protocol during the final phase. Failure to sign is attributable to specific inputs and therefore can be mitigated by the coordinator. This allows the remaining honest participants to restart the protocol. Denial of service is not costless because unspent transaction outputs are a limited resource.
+Signatures can only be made after a transaction has been negotiated, and all inputs must provide a valid signature. Consequentially users can always disrupt the protocol during the final phase. Failure to sign is attributable to specific inputs and therefore can be punished by the coordinator. This allows the remaining honest participants to restart the protocol. Denial of service is not costless because unspent transaction outputs are a limited resource.
```
nothingmuch, line 584: i prefer mitigated, it's broader and avoids going into detail about what punishment consists of, the point is that malicious users can be prevented from further disruption so attacks on availability are limited in their practicality

MaxHillebrand:
```suggestion
Signatures can only be made after a transaction has been negotiated, and all inputs must provide a valid signature. Consequentially users can always disrupt the protocol during the final phase. Failure to sign is attributable to specific inputs and therefore can be mitigated by the coordinator. This allows the remaining honest participants to restart the protocol. Denial of service is not costless because unspent transaction outputs are a limited resource.
```

PR #76, opened 2020-07-12
network identity state diagram

nothingmuch:


nopara73:
> can you please provide a screenshot of the rendered diagram?

image

nothingmuch:
weird, i'm not getting any errors locally or on planttext.com,

https://github.com/zkSNACKs/WabiSabi/blob/48060c39e27ed8281c535e16b4b917212a574e28/protocol.md#network-identity-state-diagram

image

protocol.md
```diff
@@ -192,3 +192,15 @@ The following shows the coordinator's state transitions pertaining to a
...
+
+## Network Identity State Diagram
+
+The following diagram shows the client's state transitions for a specific
+circuit. Once a circuit reaches the terminal state the identity must no longer
+be used. If a connection is broken a new circuit may be used to submit
+subsequent requests related to the same input (connection confirmations,
+signature). Therefore at most one input or output should be associated with a
+single network identity to avoid privacy leaks, but multiple network identities
+may be associated with a single input or output.
```
nopara73, line 204: Make it tech agnostic. Also I think stream would be the right word to use: https://tor.stackexchange.com/questions/16398/signal-newnym-vs-stream-isolation

```suggestion
The following diagram shows the client's state transitions for a specific
identity. Once an identity reaches the terminal state the identity must no longer
be used. If a connection is broken a new identity may be used to submit
subsequent requests related to the same input (connection confirmations,
signature). Therefore at most one input or output should be associated with a
single network identity to avoid privacy leaks, but multiple network identities
may be associated with a single input or output.
```

nothingmuch: i found it a bit strange to say the identity has a state, but reading it now it seems fine

PR #78, opened 2020-09-05
Typo

nopara73:


nothingmuch:
oops, i'm an idiot, i didn't notice this is to the .md file which is just to get SVG previews in github, the source file is .org

i'm going to rebase this in my branch to apply to the right file

PR #85, opened 2020-10-19
Formatting + Bring forward round param sig check

nopara73:
Changes from least significant to most significant:

1. I did some minor formatting for consistency, like I replaced the C#-ish `//` and used a `-` instead, or that I consistently used `sig` for `signature` and `param` for `parameter`.
2. I renamed `CoordParams` to `RoundParams`, because that's how we refer to it later on.
3. Finally I brought the `RoundParamsSig` giveout forward from signing phase to connection confirmation phase. The reason is, because this way we don't give out the round param sigs along with the large coinjoin, but rather at connection confirmation, so the data flow is more balanced. But more importantly if the user gets into issues verifying these signatures (like RPC call to Bitcoin Core stucks or something like that) then the round doesn't get DoS attacked by it. Although the Alice will still be punished, the round can progress forward just fine.

### DIFF
--- a/protocol.md
+++ b/protocol.md
@@ -124,30 +124,31 @@ The Interaction Diagram denotes Satoshi actors those are polling the Coordinator
 - `Phase`
 - `RoundId`
 - `PhaseDeadline`
-- `BlameOf` // If it's a blame round, then it is the `RoundId` of the parent round.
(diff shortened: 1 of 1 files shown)


nothingmuch (approved):
LGTM.

There's a few other potential minor changes that I want to bring up this week as we discuss tx structure

PR #86, opened 2020-11-06
Fix pronoun in Input Registration

benthecarman:
It says `She` but it is referring to `the user` which has no gender.

### DIFF
--- a/main.tex
+++ b/main.tex
@@ -222,7 +222,7 @@ For fault tolerance, request handling should be idempotent, allowing a client to
 \end{figure}
 
 The user submits an input of amount $a_{\mathit{in}}$ along with $k$ group attributes, $(M_{a_i})$.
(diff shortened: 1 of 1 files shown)


MaxHillebrand (commented):
afaik, @seresistvanandras wrote `she` for the user always.

seresistvanandras:
Thanks for pointing this out! With @nothingmuch  we are in the middle of a major review/rewriting phase. We'll definitely fix this error as well. Hopefully, soon we can release a more complete version of the writeup.

nothingmuch:
as discussed on slack, closing this because:

1. it's a well established convention
2. doesn't really hurt understanding, in fact it can introduce ambiguity
2. lots of time was already lost proof reading due to inconsistencies due to multiple rounds of revision (before the user was both Alice and Bob, and when that was removed in many places "it" was incorrectly used instead)

PR #89, opened 2020-12-30
Amount org notebook

nothingmuch:
This notebook attempts to document some of the C# work we did and the justifications behind it, as well as try new things, and explain them to a broader audience using many plots.

This PR does not yet contain the original goal of this, which was twofold

1. simulate UTXO set management to see effects on fragmentation
2. examine cost function in detail

However this is already a substantial amount of text, so I thought this was a good chunk of work to add as a PR.


nopara73:
Thank you for capitalizing properly. Would you mind to explain what you're trying to do here?

nothingmuch:
in this pr:

1. explain the ideas behind the things that we tried
2. compare alternatives more systematically
3. address the probable denominations question

and not yet included, since i'm still working on it:

1. the knapsack based optimization i've described
2. defining a complete cost function as @seresistvanandras and i have been discussing for a few months
3. agent based approach to simulate mix graphs and evaluate mix graph topology, blockspace and utxo fragmentation

nopara73:
If this is just a specification, then why not just create an `.md` file that anyone can read and contribute to?

nothingmuch:
It's not a specfication, it's a bunch of experiments similar to the C# ones but focusing on some other areas. As for why not md, because it's not just text, the code and the plots are the bulk of it

guix/README.md
```diff
@@ -0,0 +1,19 @@
+# Reproducible Notebook with Guix
```
nopara73, line 1: Please make this section better.

nopara73 (changes requested):
Can't get started.

guix/README.md
```diff
@@ -0,0 +1,79 @@
+# Jupyter Notebook Dependencies
+
+Although GitHub renders Jupyter notebooks, in order to be able to use the
+interactive portions or to run/modify the code yourself, you will need the
+python dependencies.
```
nopara73, line 5: Thanks, much better. For example I wasn't aware of this crucial fact:

> GitHub renders Jupyter notebooks

With this in mind, can you create another file for this setup and leave only the link to the `Amount Org.ipynb` in the README?

nopara73 (changes requested):
Much better, thank you! Will review the content later.

guix/README.md
nothingmuch: Sorry I figured you'd just be [reading on github](https://github.com/zkSNACKs/WabiSabi/blob/amount-org-notebook/Amount%20Org.ipynb) to start so I didn't take much care.

I expanded the instructions a bit, I hope it's enough but i'm not sure if this will work under WSL and I've got no windows machine to try.

If you are interested in trying out guix (highly reccomended IMO) I'm happy to help get started with it but it shouldn't really be a requirement, it's just my personal preference and happens to support reproducibility well so I like using it for dev environments

nopara73 (changes requested):
One cannot comment on `Amount Org.ipynb` as the file is too large and the content doesn't load, so basically it's impossible to properly review. I'm not sure how to progress forward. I can go though it alone, but I won't be able to comment.

guix/README.md
nothingmuch: you mean merge the `guix/README.md` file into the existing `README`? maybe more organized would be to put the notebook stuff in a separate subdirectory? i didn't really think about the layout but in hindsight it was probably a mistake to clutter up the root directory

nothingmuch:
Oh, damn, I thought comitting the source and output data separately would be enough. I can break it up into smaller commits/PRs so that they can be reviewed individually

nopara73:
Maybe break it down for the comments and before merge we put them back into one doc?

nothingmuch:
That won't work since there are dependencies between the cells, i can re-commit it in several pieces so that each diff is smaller, hopefully that'll convince github to show something

nopara73:
We could also just merge this and work on this as a basis.

nothingmuch:
reviewnb appears to be pretty usable, Lucas enabled it for the repository earlier today :

https://app.reviewnb.com/zkSNACKs/WabiSabi/pull/89/files/

As far as I can tell it posts normal review comments back to github, but i can't find a way to make annotations on individual lines, only the cells. I just pushed an ammended version of the commits with some mistakes fixed and the larger cells split up, but there are still a few larger cells that might be hard to review

nopara73 (changes requested):
Write this in a traditional `.md` file and don't create unnecessary obstacles like installing buggy linuxy tools and make reviewers to give permissions to random sites on the Internet. This is no way to work. Please respect the time of others.

PR #95, opened 2021-01-27
Final CES submission

nothingmuch:
Apologies for the super long delay in bringing this back in sync.

This PR contains all of the modifications made to the submitted version of the paper, and includes changes based on the reviewers' excellent feedback, which covers many minor fixes and a few major ones, such as more explicit system and threat models.


nothingmuch:
force push just adds co-authored-by

main.tex
```diff
@@ -50,99 +61,204 @@
...
 \end{abstract}
 
 \section{Introduction}
+Bitcoin transactions transfer funds by consuming unspent outputs of previous transactions as inputs to create new outputs~\cite{nakamoto2009bitcoin}. The protocol rules enforced by the network ensure that transactions do not arbitrarily inflate the money supply and that outputs are spent at most once. While some newer cryptocurrencies use more sophisticated approaches to define such rules, in Bitcoin the amounts, as well as the specific outputs being spent, are broadcast in the clear as part of the transaction. This presents significant challenges to transacting privately\footnote{In this work we restrict the discussion of Bitcoin privacy to that of public ledger transactions, but there are other considerations especially at the network layer. For a more comprehensive discussion see \url{https://en.bitcoin.it/wiki/Privacy}.} as shown already in some of the earliest academic studies of Bitcoin~\cite{reid2013analysis,ron2013quantitative,androulaki2013evaluating,ober2013structure,moeser2013inquiry,meiklejohn2013fistful}.
```
MaxHillebrand, line 87: double of `transactions` and `transfer` is not that beautiful. maybe:
```suggestion
Bitcoin transactions reassign funds by consuming unspent outputs of previous transactions as inputs to create new outputs~\cite{nakamoto2009bitcoin}. The protocol rules enforced by the network ensure that transactions do not arbitrarily inflate the money supply and that outputs are spent at most once. While some newer cryptocurrencies use more sophisticated approaches to define such rules, in Bitcoin the amounts, as well as the specific outputs being spent, are broadcast in the clear as part of the transaction. This presents significant challenges to transacting privately\footnote{In this work we restrict the discussion of Bitcoin privacy to that of public ledger transactions, but there are other considerations especially at the network layer. For a more comprehensive discussion see \url{https://en.bitcoin.it/wiki/Privacy}.} as shown already in some of the earliest academic studies of Bitcoin~\cite{reid2013analysis,ron2013quantitative,androulaki2013evaluating,ober2013structure,moeser2013inquiry,meiklejohn2013fistful}.
```

main.tex
```diff
@@ -50,99 +61,204 @@
...
+A fundamental requirement for electronic money is double-spending prevention, and Bitcoin's main innovation was preventing double-spending and illegal inflation without relying on a trusted authority, thereby disintermediating transactions. This is in contrast to online e-cash schemes where a server authorizes transactions or offline schemes where the identity of the double spender is revealed allowing the authority to intervene after the fact. Bitcoin's relative success in comparison suggests that the lack of trusted third parties factored more strongly in its adoption than the comparatively stronger privacy guarantees provided by the (possibly revocable) transaction  anonymity traditionally associated with e-cash~\cite{davida1997anonymity}.
```
MaxHillebrand, line 89: `Digital Bearer Certificate [DBC]` is afaik the neutral term for these things. `e-Cash` was an early implementation of a DBC by Chaum.

```suggestion
A fundamental requirement for electronic money is double-spending prevention, and Bitcoin's main innovation was preventing double-spending and illegal inflation without relying on a trusted authority, thereby disintermediating transactions. This is in contrast to digital bearer certificates schemes where a server authorizes transactions or offline schemes where the identity of the double spender is revealed allowing the authority to intervene after the fact. Bitcoin's relative success in comparison suggests that the lack of trusted third parties factored more strongly in its adoption than the comparatively stronger privacy guarantees provided by the (possibly revocable) transaction  anonymity traditionally associated with e-cash~\cite{davida1997anonymity}.
```

main.tex
```diff
@@ -50,99 +61,204 @@
...
+Chaumian CoinJoin~\cite{mizrahi2013blind,maxwell2013coinjoin,zerolink} is a privacy-enhancing technique that uses the atomicity of transactions and Chaumian blind signatures~\cite{chaum1983blind} to construct collaborative Bitcoin transactions, also known as CoinJoins. Participants connect to a server, known as the \emph{coordinator}, and submit their inputs and outputs using different anonymity network identities. That alone would provide anonymity but since outputs are unconstrained it's not robust against malicious users who may disrupt the protocol by claiming more than their fair share in the transaction outputs. To mitigate this the coordinator provides blind signatures representing units of standard denominations in response to submitted inputs. Clients unblind and submit the output with valid signature, which allows the to verify that an output registration is authorized without being able to link it to the corresponding inputs.
```
MaxHillebrand, line 97:
```suggestion
Chaumian CoinJoin~\cite{mizrahi2013blind,maxwell2013coinjoin,zerolink} is a privacy-enhancing technique that uses the atomicity of transactions and Chaumian blind signatures~\cite{chaum1983blind} to construct collaborative Bitcoin transactions, also known as CoinJoins. Participants connect to a server, known as the \emph{coordinator}, and submit their inputs and outputs using different anonymity network identities. That alone would provide anonymity but since outputs are unconstrained it's not robust against malicious users who may disrupt the protocol by claiming more than their fair share in the transaction outputs. To mitigate this the coordinator provides blind signatures representing units of standard denominations in response to submitted inputs. Clients unblind and submit the output with valid signature, which allows them to verify that an output registration is authorized without being able to link it to the corresponding inputs.
```

main.tex
```diff
@@ -50,99 +61,204 @@
...
+
+WabiSabi can be instantiated to construct a variety of CoinJoin transaction structures that depart from the standard output denomination convention, as in SharedCoin\footnote{See: \url{https://github.com/sharedcoin/Sharedcoin}} and CashFusion\footnote{See: \url{https://github.com/cashshuffle/spec}} style transactions and Knapsack~\cite{maurer2017anonymous} mixing. Payments from CoinJoin transactions are possible which enhances sender privacy, as are payments within CoinJoin transactions, improving privacy for both sender and receiver and potentially hiding the payment amount. Additionally, restrictions on the consolidation of inputs can be removed, and there are opportunities for reducing unmixed change, relaxing minimum required denominations, and improved block space efficiency compared to what is possible with blind signatures.
```
MaxHillebrand, line 105: do we want to mention that payments from/within CJ are a substantial fee savings, as it is a specific case of transaction batching. 

```suggestion
WabiSabi can be instantiated to construct a variety of CoinJoin transaction structures that depart from the standard output denomination convention, as in SharedCoin\footnote{See: \url{https://github.com/sharedcoin/Sharedcoin}} and CashFusion\footnote{See: \url{https://github.com/cashshuffle/spec}} style transactions and Knapsack~\cite{maurer2017anonymous} mixing. Payments from CoinJoin transactions are possible which enhances sender privacy, as are payments within CoinJoin transactions, improving privacy for both sender and receiver and potentially hiding the payment amount. Both payments from and within CoinJoin transactions benefit from the transaction size savings of payment batching. Additionally, restrictions on the consolidation of inputs can be removed, and there are opportunities for reducing unmixed change, relaxing minimum required denominations, and improved block space efficiency compared to what is possible with blind signatures.
```

main.tex
```diff
@@ -50,99 +61,204 @@
...
 
-The use of standard denominations in the resulting CoinJoin transaction obscures the relationship between individual inputs and outputs, making the origins of each output ambiguous. Unfortunately standard denominations limit the use of privacy-enhanced outputs for payments of arbitrary amounts and result in a change output which maintains a link to the non-private input.
+\subsubsection{Wasabi's ZeroLink Protocol}
+
+Several CoinJoin protocols have been implemented and many more proposed, with varying privacy and trust models. For an overview see~\cref{sec:relatedwork}. ZeroLink~\cite{zerolink} is one such protocol implemented by Wasabi, the most popular Chaumian CoinJoin implementation for Bitcoin. Although the coordinator must be trusted to ensure availability, but no trust is required with regards to safeguarding users' funds or keeping mapping between inputs and outputs private. The protocol breaks down into three main phases:
```
MaxHillebrand, line 172: The coordinator is trusted with input/input linkage and input/change output linkage.

```suggestion
Several CoinJoin protocols have been implemented and many more proposed, with varying privacy and trust models. For an overview see~\cref{sec:relatedwork}. ZeroLink~\cite{zerolink} is one such protocol implemented by Wasabi, the most popular Chaumian CoinJoin implementation for Bitcoin. Although the coordinator must be trusted to ensure availability, but no trust is required with regards to safeguarding users' funds or keeping mapping between inputs and equal amount outputs private. The protocol breaks down into three main phases:
```

main.tex
```diff
@@ -50,99 +61,204 @@
...
+
+\begin{description}
+ \item[Input Registration] Users register their inputs along with proofs (digital signatures) that they own those inputs. Additionally, users provide their change outputs and blinded outputs to the coordinator. The coordinator blindly signs the requested blinded output.
```
MaxHillebrand, line 175:
```suggestion
 \item[Input Registration] Users register their inputs along with proofs (digital signatures) that they can spend those inputs. Additionally, users provide their change outputs and blinded outputs to the coordinator. The coordinator blindly signs the requested blinded output.
```

main.tex
```diff
@@ -50,99 +61,204 @@
...
...ool\footnote{See: \url{https://www.samouraiwallet.com/whirlpool}}, and other apparent CoinJoin transactions are provided. The ``Other'' category includes JoinMarket\footnote{See: \url{https://github. com/JoinMarket-Org/joinmarket}}, but also has an inherent false positive error given these transactions are identified heuristically. We collected every equal output CoinJoin-like transactions from the Bitcoin blockchain till 2020 May 22nd. Wasabi and Samourai CoinJoin transactions are trivial to detect. A transaction that has more than $10$ equal amount outputs are surely Wasabi CoinJoins, while a Samourai CoinJoin transaction has the same structure, with $5$ equal amount outputs of a fixed size, $3$ inputs of exactly the same amount, while fees are paid by $2$ inputs of a slightly higher amount which are direct descendants of a so called ``tx0'', a transaction that prepares outputs for mixing and pays the coordination fees. However, other CoinJoin transactions, e.g. JoinMarket, are more difficult to identify~\cite{moeser2017anonymous}. Heuristically, we deem every transaction being in the ``Other'' CoinJoin transaction category if they simultaneously satisfy the following conditions:
```
MaxHillebrand, line 181: @nopara73 ran the dumplings repo again, should this be updated here?
I'm not sure on the correct date though. 

```suggestion
In this work, we aim to improve on the ZeroLink protocol as implemented in Wasabi. We identify several privacy shortcomings and inefficiencies of Wasabi CoinJoins. Specifically, we are interested in quantifying block-space and other inefficiencies stemming from the imposed minimum denomination. These metrics comparing Wasabi, Samourai's Whirlpool\footnote{See: \url{https://www.samouraiwallet.com/whirlpool}}, and other apparent CoinJoin transactions are provided. The ``Other'' category includes JoinMarket\footnote{See: \url{https://github. com/JoinMarket-Org/joinmarket}}, but also has an inherent false positive error given these transactions are identified heuristically. We collected every equal output CoinJoin-like transactions from the Bitcoin blockchain till 2021 Jan. Wasabi and Samourai CoinJoin transactions are trivial to detect. A transaction that has more than $10$ equal amount outputs are surely Wasabi CoinJoins, while a Samourai CoinJoin transaction has the same structure, with $5$ equal amount outputs of a fixed size, $3$ inputs of exactly the same amount, while fees are paid by $2$ inputs of a slightly higher amount which are direct descendants of a so called ``tx0'', a transaction that prepares outputs for mixing and pays the coordination fees. However, other CoinJoin transactions, e.g. JoinMarket, are more difficult to identify~\cite{moeser2017anonymous}. Heuristically, we deem every transaction being in the ``Other'' CoinJoin transaction category if they simultaneously satisfy the following conditions:
```

main.tex
```diff
@@ -50,99 +61,204 @@
...
 
-We define \emph{CoinJoin inefficiency} as the fraction of non-mixed change outputs in a CoinJoin transaction, see \cref{fig:cjinefficiency}.
+We define \emph{CoinJoin inefficiency} as the fraction of non-mixed change outputs in a CoinJoin transaction. In~\cref{fig:cjinefficiency}, we observe the large amount of non-mixed outputs of essentially all Bitcoin privacy-enhancing tools. Samourai has the lowest CoinJoin inefficency, since it always applies $5$ equal amount of inputs and outputs. We note a sharp decline of CoinJoin inefficiency in Wasabi CoinJoins around January 2019 due to the introduction of multiples of the base denomination.
```
MaxHillebrand, line 197: Note that in samourai there are not `5 equal amount of inputs`, there are two different amounts, represented in 2 and 3 coins.

main.tex
```diff
@@ -50,99 +61,204 @@
...
     \label{fig:postmixmerging}
+\end{minipage}
 \end{figure}
 
-\subsubsection{Lack of privacy-enhanced payments} Currently Wasabi supports neither payments from a CoinJoin, nor payments in a CoinJoin. Payments from a CoinJoin would protect sender privacy and improve efficiency by requiring fewer intermediate outputs. Payments within a CoinJoin would protect both sender and receiver privacy, and since they are a form of PayJoin\footnote{\url{https://en.bitcoin.it/wiki/PayJoin}} it would also improve privacy by introducing degrees of freedom in the interpretation of CoinJoins.
-
-\subsection{Our Contribution}
+\subsubsection{Lack of privacy-enhanced payments} Currently, Wasabi supports neither payments from a CoinJoin, nor payments in a CoinJoin. Payments from a CoinJoin would protect sender privacy and improve efficiency by requiring fewer intermediate outputs. Payments within a CoinJoin would protect both sender and receiver privacy, as well as undermine clustering heuristics since they are analogous to PayJoins. It would also improve privacy by introducing degrees of freedom in the interpretation of CoinJoins.
```
MaxHillebrand, line 254: @nothingmuch you are the first I know to say `payment from CoinJoin` and `payment within CoinJoin`. 
is this sufficiently explained here?

main.tex
```diff
@@ -174,41 +290,83 @@ \subsection{Zero-knowledge proofs of knowledge}
...
+    \item[Bitcoin P2P network nodes] Typically users do not run Bitcoin full nodes. Therefore, they rely on other full nodes to serve them necessary data from the Bitcoin blockchain. For instance, users invoke full nodes whenever they verify the inclusion of a CoinJoin transaction in the blockchain or query their own balance. Note, that for the sake of privacy, these queries needs to be made unlinkable as well.
+\end{description}
+
+\subsection{Communication model} \label{sec:commodel}
+We assume all messages (onion-routed messages, broadcast transactions, queries about the blockchain) are delivered with a maximum delay under the bounded synchronous communication setting~\cite{attiya2004distributed}. Furthermore, we assume all actors
```
MaxHillebrand, line 312: fwiw, this part has line breaks for text-blocks, the above parts did not have. 

not sure if this is intended.

main.tex
```diff
@@ -174,41 +290,83 @@ \subsection{Zero-knowledge proofs of knowledge}
...
+In order to facilitate construction of a CoinJoin transaction while protecting the privacy of participants, we instantiate the scheme with a single group attribute $M_a$ which encodes a confidential Bitcoin amount as a Pedersen commitment. These commitments are never opened. Instead, properties of the values they commit to are proven in zero-knowledge, allowing the coordinator to validate requests made by honest participants. In ideal circumstances, the coordinator would not learn anything beyond what can be learned from the resulting CoinJoin transaction but despite the unlinkability of the credentials timing of requests or connectivity issues may still reveal information about links.
```
MaxHillebrand, line 353: `B` the protocol, `b` the currency.

```suggestion
In order to facilitate construction of a CoinJoin transaction while protecting the privacy of participants, we instantiate the scheme with a single group attribute $M_a$ which encodes a confidential bitcoin amount as a Pedersen commitment. These commitments are never opened. Instead, properties of the values they commit to are proven in zero-knowledge, allowing the coordinator to validate requests made by honest participants. In ideal circumstances, the coordinator would not learn anything beyond what can be learned from the resulting CoinJoin transaction but despite the unlinkability of the credentials timing of requests or connectivity issues may still reveal information about links.
```

main.tex
nothingmuch: in the literature being cited here "online e-cash" and "offline e-cash" are standard terms, i think DBC only came about after the blockchain fad started

main.tex
nothingmuch: "reassign" is technically accurate but the paragraph is meant to give minimum background to people who aren't familiar with the UTXO model, so i think transfer is more intuitive to that specific target audience

main.tex
nothingmuch: this is an important detail, thanks

main.tex
nothingmuch: this shouldn't be updated unless we also update all of the charts... @seresistvanandras does this need the raw data files? how much work is it?

main.tex
nothingmuch: ack. also @seresistvanandras i think you need to stop using "apply" as a catch all verb ;-)

main.tex
nothingmuch: this is explained in more detail in the examples section, i remember working on this paragraph and giving up trying to explain the difference precisely before the concept of credentials was introduced

main.tex
nothingmuch: i think we used uppercase for both throughout, and possibly also "satoshi amount" when emphasizing that it's an integer value. Maybe "BTC" is better for the currency?

at any rate if we change this we should make sure we go through the entire doc to make sure it's consistent

main.tex
```diff
@@ -50,99 +61,204 @@
...
+Chaumian CoinJoin~\cite{mizrahi2013blind,maxwell2013coinjoin,zerolink} is a privacy-enhancing technique that uses the atomicity of transactions and Chaumian blind signatures~\cite{chaum1983blind} to construct collaborative Bitcoin transactions, also known as CoinJoins. Participants connect to a server, known as the \emph{coordinator}, and submit their inputs and outputs using different anonymity network identities. That alone would provide anonymity but since outputs are unconstrained it's not robust against malicious users who may disrupt the protocol by claiming more than their fair share in the transaction outputs. To mitigate this the coordinator provides blind signatures representing units of standard denominations in response to submitted inputs. Clients unblind and submit the output with valid signature, which allows them to verify that an output registration is authorized without being able to link it to the corresponding inputs.
```
nothingmuch, line 97: this is not a typo but a word-o ;-)

```suggestion
Chaumian CoinJoin~\cite{mizrahi2013blind,maxwell2013coinjoin,zerolink} is a privacy-enhancing technique that uses the atomicity of transactions and Chaumian blind signatures~\cite{chaum1983blind} to construct collaborative Bitcoin transactions, also known as CoinJoins. Participants connect to a server, known as the \emph{coordinator}, and submit their inputs and outputs using different anonymity network identities. That alone would provide anonymity but since outputs are unconstrained it's not robust against malicious users who may disrupt the protocol by claiming more than their fair share in the transaction outputs. To mitigate this the coordinator provides blind signatures representing units of standard denominations in response to submitted inputs. Clients unblind and submit the output with valid signature, which allows the coordinator to verify that an output registration is authorized without being able to link it to the corresponding inputs.
```

nopara73 (approved):
Since this PR had been reviewed already outside of GitHub I suggest to merge it after @MaxHillebrand is happy.

MaxHillebrand (approved):
I'd like these to be fixed:

- https://github.com/zkSNACKs/WabiSabi/pull/95#discussion_r567669641
- https://github.com/zkSNACKs/WabiSabi/pull/95#discussion_r567663524 [trivial]
- https://github.com/zkSNACKs/WabiSabi/pull/95#discussion_r567666065
- https://github.com/zkSNACKs/WabiSabi/pull/95#discussion_r567669641 [important]

But other than these nits, ACK

main.tex
lontivero, line 89: The word legal is confusing to me, I mean, I understand what it means in this context but i don't think it is okay. I was not able to find a better word but this is my suggestion:

```suggestion
A fundamental requirement for electronic money is double-spending prevention, and Bitcoin's main innovation was preventing double-spending and controlling supply without relying on a trusted authority, thereby disintermediating transactions. This is in contrast to online e-cash schemes where a server authorizes transactions or offline schemes where the identity of the double spender is revealed allowing the authority to intervene after the fact. Bitcoin's relative success in comparison suggests that the lack of trusted third parties factored more strongly in its adoption than the comparatively stronger privacy guarantees provided by the (possibly revocable) transaction  anonymity traditionally associated with e-cash~\cite{davida1997anonymity}.
```

main.tex
```diff
@@ -50,99 +61,204 @@
...
+
+\subsection{Privacy issues and CoinJoin transactions}
+
+Bitcoin is a pseudonymous cryptocurrency~\cite{nakamoto2009bitcoin}: coins are associated with so called addresses, which encode the spending conditions of the coin (their \texttt{scriptPubKey}). Bitcoin transactions spend coins as inputs and create new coins output in place of the spent ones. The Bitcoin protocol mandates the rules of a valid transaction (no double-spending, valid signatures, etc.), which is verified by all nodes of the network. Since every transaction is recorded in a public ledger, the flow of money is traceable. Several idioms of use allow one to cluster addresses likely to be controlled by the same user. Many heuristics were devised in previous work~\cite{meiklejohn2013fistful,androulaki2013evaluating,reid2013analysis,ron2013quantitative}. The most commonly used heuristic to cluster Bitcoin addresses is the \emph{common input ownership heuristic}. It states that in any Bitcoin transaction, all the inputs are likely controlled by the same user. This heuristic already facilitates a fine-grained clustering of Bitcoin addresses, with alarming accuracy\cite{nick2015data}.
```
lontivero, line 115: I think the word `alarming` makes sense to us because be believe it is alarming but I am not sure that word is the best for a research paper because it could sound as sensationalism.

```suggestion
Bitcoin is a pseudonymous cryptocurrency~\cite{nakamoto2009bitcoin}: coins are associated with so called addresses, which encode the spending conditions of the coin (their \texttt{scriptPubKey}). Bitcoin transactions spend coins as inputs and create new coins output in place of the spent ones. The Bitcoin protocol mandates the rules of a valid transaction (no double-spending, valid signatures, etc.), which is verified by all nodes of the network. Since every transaction is recorded in a public ledger, the flow of money is traceable. Several idioms of use allow one to cluster addresses likely to be controlled by the same user. Many heuristics were devised in previous work~\cite{meiklejohn2013fistful,androulaki2013evaluating,reid2013analysis,ron2013quantitative}. The most commonly used heuristic to cluster Bitcoin addresses is the \emph{common input ownership heuristic}. It states that in any Bitcoin transaction, all the inputs are likely controlled by the same user. This heuristic already facilitates a fine-grained clustering of Bitcoin addresses, with high accuracy\cite{nick2015data}.
```

nothingmuch:
> I'd like these to be fixed:

first and last items are duplicate, did a 4th go missing?

second item is superseded by my correction (should be "the coordinator" not "them")

main.tex
lontivero: Alternatively to Max's suggestion it could be `control`

main.tex
```diff
@@ -50,99 +61,204 @@
...
-    \includegraphics[scale=0.4]{Figures/SmallValueInputsWasabi.pdf}
+    \includegraphics[width=\linewidth]{Figures/SmallValueInputsWasabi.pdf}
     \caption{Fraction of inputs with value smaller than the minimum denomination in Wasabi CoinJoin transactions.}
     \label{fig:minimumdenomination}
+\end{minipage}
+\hspace{6pt}
+\begin{minipage}{.45\textwidth}
+    \centering
+    \includegraphics[width=0.95\linewidth]{Figures/freshBitcoins.pdf}
+    \caption{Volume of unmixed coins entering to various CoinJoin schemes. Wasabi is the most popular Chaumian CoinJoin implementation.}
```
lontivero, line 222: Here we use `unmixed` while in other places we call them `non-mixed`. Is this okay?

main.tex
lontivero: This is clarified below: 
> the coordinator learns that those inputs belong to the same user

It could be redundant to repeat it here.

main.tex
```diff
@@ -281,21 +434,27 @@ \subsubsection{Unified Registration}\label{sec:unified}
...
+To prevent the coordinator from being able to distinguish between initial vs. subsequent input registration requests (which may merge amounts) credential presentation should be mandatory. Initial credentials can be obtained with an auxiliary bootstrapping operation (\cref{fig:bootstrap}). A more sophisticated approach would be to prove knowledge of a valid MAC or that the amount attribute has value a value of zero, but because the input registration phase starts synchronously when enough users have indicated their intent to participate this would not actually save a round trip in practice.
 
 \subsection{Signing phase}
 
-The user fetches the finalized but unsigned transaction from the coordinator. If it contains the outputs she registered she will sign her inputs and submit each signature separately using the network identity used for that input's registration.
+The user fetches the finalized but unsigned transaction from the coordinator. If it contains her registered output she will sign her inputs and submit each signature separately using the network identity used for the corresponding input's registration.
```
lontivero, line 441: It should say outputs i think because t is more generic.
```suggestion
The user fetches the finalized but unsigned transaction from the coordinator. If it contains her registered outputs she will sign her inputs and submit each signature separately using the network identity used for the corresponding input's registration.
```

main.tex
```diff
@@ -281,21 +434,27 @@ \subsubsection{Unified Registration}\label{sec:unified}
...
 
 \subsection{Examples}\label{sec:examples}
 
-% todo sub figures
+To illustrate the above protocols, \cref{fig:ex1,fig:ex2} show how a user might register inputs and outputs with the simplified protocols, where credentials are only requested at input registration and presented at output registration. \Cref{fig:ex3,fig:ex4} show the unified protocol, when credentials are both presented and requested in every request.
```
lontivero, line 445:
```suggestion
To illustrate the above protocols, \cref{fig:ex1,fig:ex2} shows how a user might register inputs and outputs with the simplified protocols, where credentials are only requested at input registration and presented at output registration. \Cref{fig:ex3,fig:ex4} shows the unified protocol, when credentials are both presented and requested in every request.
```

nothingmuch: `\cref` knows how to pluralize when given multiple arguments, so this renders as "figs. 11a and 11b show..."

main.tex
```diff
@@ -281,21 +434,27 @@ \subsubsection{Unified Registration}\label{sec:unified}
   \label{fig:bootstrap}
 \end{figure}
 
-To prevent the coordinator from being able to distinguish between initial vs. subsequent input registration requests (which may merge amounts) credential presentation should be mandatory. Initial credentials can be obtained with an auxiliary bootstrapping operation (\cref{fig:bootstrap}).
+To prevent the coordinator from being able to distinguish between initial vs. subsequent input registration requests (which may merge amounts) credential presentation should be mandatory. Initial credentials can be obtained with an auxiliary bootstrapping operation (\cref{fig:bootstrap}). A more sophisticated approach would be to prove knowledge of a valid MAC or that the amount attribute has value a value of zero, but because the input registration phase starts synchronously when enough users have indicated their intent to participate this would not actually save a round trip in practice.
```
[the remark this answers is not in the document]
nothingmuch: in dyslexic english this is perfectly natural ;-)

main.tex
nothingmuch: to me non-mixed implies something more intrinsic than unmixed, the latter kind of suggests that coins can start unmixed and then become mixed which is technically not true, so I think that's how we ended up with that, i think consistency is nice, but we should probably ask a native speaker if either of them sounds weird and if the connotation i'm getting is just in my head... @DanGould, do you have any thoughts?

main.tex
nothingmuch: "but" is also not grammatical after "Although" already used

main.tex
nothingmuch: both of these are better than "own" (although we say what "ownership" means above so it's at least well defined), i think i like "control" the best but not sure

main.tex
lontivero: wow. i didn't know that.

main.tex
```diff
@@ -448,7 +604,7 @@ \subsection{Credential Requests}
 For each $i \in [1, k]$ the user chooses an amount $a_i \mid 0 \leq a_i < a_{\mathit{max}}$ subject to the constraints of the balance proof (\cref{balance}). She commits to the amount with randomness $r_i \in_R \mathbb{Z}_q$, and these commitments are the attributes of the requested credentials:
 \[ M_{a_i}={G_h}^{r_i}{G_g}^{a_i} \]
 
-For each amount $a_i$ she also computes a range proof which ensures there are no negative values:
+For each amount $a_i$ she also computes a range proof which ensures the amounts are in the allowed range which ensures that finite field arithmetic will not overflow and the values will behave like positive integers when added or subtracted:
```
lontivero, line 607: > which ensures the amounts are in the allowed range which ensures

`which ensures` repeats twice in the same sentence.

main.tex
```diff
@@ -578,105 +732,156 @@ \subsubsection{Malicious Coordinator}
...
+We can model registration requests as vertices of a directed acyclic graph labeled by $\Delta_a$ and credentials as edges labelled by the amount in the attribute, connecting the registration request where a credential was issued to where it was presented, much like the graph depicted in \cref{sec:examples}. Apart from the bootstrap requests which have indegree 0 (no credentials are presented) and final output requests which have outdegree 0 (requested credentials are not presented in any subsequent request), this graph is $k$-regular. The serial number and balance proofs enforce a global invariant where the sums of the inwards and outwards edges of every vertex differ by its label, and so long as honest participants are able to make their registrations it will be balanced (both vertex and edge labels will sum to 0).
```
lontivero, line 749: > (requested credentials are not presented in any subsequent request)

Why not? I think it could be possible to request credentials during output registration and present them in subsequent requests.

main.tex
```diff
@@ -578,105 +732,156 @@ \subsubsection{Malicious Coordinator}
...
+The unlinkability of credentials and hiding property of the amount commitments obscure the edges and their labels from the coordinator and other participants. However, the coordinator observes registration requests according to a partial order which is an extension of the partial order defined by the DAG, in other words, the coordinator knows that parallel requests do not share an edge and that dependent requests must be made sequentially. In other words information about the order and timing of requests as well as the inherently overt transaction data relayed in the registration requests can constrain the graph topology allowing the coordinator to draw inferences about the edges despite not being able to observe the edge set directly.
```
lontivero, line 751: `in other words` repeats twice and if well they are not in the same sentence they are close enough.

main.tex
```diff
@@ -578,105 +732,156 @@ \subsubsection{Malicious Coordinator}
...
+
+\subsection{Theft prevention}
+
+Since the output of the protocol is a single Bitcoin transaction theft is prevented by only signing the transaction if the outputs are as expected, so the protocol inherits theft resistance directly from Bitcoin's security model.
+
+A malicious user could claim more funds than she registered if and only if she is able to forge the coordinator's MAC on some credential. This is certainly not possible unless credentials can be forged under chosen message attacks, but even if that were the case this would only result in denial of service, as the honest users would refuse to sign such a transaction.
+
+\subsection{Security Proofs}
+The above outlined security and privacy guarantees directly follow from the applied KVAC scheme of~\cite{chase2019signal}. The unlinkability of WabiSabi credentials follow from the unlinkability of the underlying keyed-verification anonymous
+credential scheme. The theft prevention of the WabiSabi protocol follows from the unforgeability of the KVAC scheme. We refer the to~\cite{chase2019signal} for formal security arguments.
```
lontivero, line 779: I think the theft prevention comes from the fact that users will not sign a transaction that takes their money and from the fact that the bitcoin network will reject an inflationary transaction. I don't think the KVAC takes part in theft prevention.

main.tex
```diff
@@ -578,105 +732,156 @@ \subsubsection{Malicious Coordinator}
...
+$\#\mathbb{Z}_q$ & $-$   & $k(3n + 1)$            & $-$   & $-$ & $5k$                                             & $2$                  & $k$ & $5k$
+\end{tabular}
+
+  \caption{Communication overhead of the WabiSabi protocol. Each round trip corresponds to a single registered input and $k$ denotes the number of requested credentials. The bit-width of the range proofs is denoted as $n$, and for bootstrap requests its value is zero.}\label{tab:protocolefficiency}
+\end{table*}
+
+\subsection{Concrete performance}
+
+A proof of concept implementation of the credential system has been added to the open source Wasabi wallet\footnote{\url{https://github.com/zkSNACKs/WalletWasabi}}. This implementation relied on \texttt{NBitcoin.Secp256k1}\footnote{\url{https://github.com/MetacoSA/NBitcoin/tree/master/NBitcoin.Secp256k1}} a port of \texttt{libsecp256k1}\footnote{\url{https://github.com/bitcoin-core/secp256k1/pull/486}}. Due to limitations of the .NET runtime only 32-bit limb field elements are supported. All proofs were made non-interactive using the strong variant of the Fiat-Shamir transform~\cite{fiat1987prove, bernhard2012not, hamburg2017strobe}.
```
lontivero, line 804: > Due to limitations of the .NET runtime only 32-bit limb field elements are supported

What are these limitations? Without a link this sounds like "Due my dog has eaten my homework".

main.tex
```diff
@@ -578,105 +732,156 @@ \subsubsection{Malicious Coordinator}
...
+
+Performance was measured on an Intel Core i7-7500U CPU 2.70GHz (Kaby Lake) using BenchmarkDotNet version 0.12.1 on Ubuntu 20.04 with .NET Core version 5.0.101. We ran a unified registration request-response flow\footnote{See: \url{https://github.com/zkSNACKs/WasabiBenchmark}}  with $k=2$ and 51 bit range proofs, and observed a $1.278$ seconds mean time with a standard deviation of $0.103$ seconds in $93$ iterations.
```
lontivero, line 806: Provably you already know this but just in case I want to clarify it: what I measured is a two rounds or unified registration request-response. This is:

* Client request bootstrap credentials.
* Coordinator issued the credentials
* Client presents the credentials and request new credentials
* Coordinator verifies the presented credentials (proofs) and issues the new credentials

I did it in this way because in the first request (requesting bootstrap credentials) there is not range proofs involved, what is the more expensive creation/verification. So, I needed to make a subsequent request, that's why the two round trips.

main.tex
```diff
@@ -578,105 +732,156 @@ \subsubsection{Malicious Coordinator}
...
  \item[Trustlessness] Some protocols have stronger trust assumptions with respect to theft, for example Mixcoin and Blindcoin\footnote{Both protocols have provisions for accountability of the mixes, mitigating theft concerns}, whereas other protocols utilize Bitcoin's scripting capabilities to secure user funds.
 \end{description}
 
-Since Chaumian CoinJoins have been successfully deployed despite some limitations (see \cref{sec:limitations}), it appears that centrally coordinated CoinJoins strike a desirable balance in the Bitcoin privacy design space. It is non-custodial, has low messaging complexity, and results in a single transaction minimizing delays and network fees.
+Additional related work on Bitcoin privacy that is based on longer lived payment channels~\cite{green2017bolt, tairi2019a2l} is less directly comparable and have been omitted. Some of the listed approaches, namely TumbleBit~\cite{heilman2017tumblebit}, BlindCoin~\cite{valenta2015blindcoin} and custodial services can be used more similarly, but here we consider their use in mixes.
```
lontivero, line 819: I never know how to apply this rule. Is it `longer lived payment channels` or `longer-lived payment channels`?

main.tex
```diff
@@ -578,105 +732,156 @@ \subsubsection{Malicious Coordinator}
...
...number of participants. The number of transactions is not necessarily comparable because different protocols have a significantly different transaction structures. Cost in terms of network fees is better provided in terms of the block weight, in terms of the minimum the number of outputs created and destroyed, with any change outputs incurring additional constant overhead. Overall wait times are also bounded by the minimum number of phases which require transactions to be confirmed in a block. Furthermore, some protocols like Chris Belcher's recent CoinSwap proposal~\cite{belcher2020design} are designed to provide robust privacy with a single iteration of the protocol, whereas others such as Mixcoin~\cite{bonneau2014mixcoin} require multiple mixes to achieve privacy. Secondly, the various protocols are qualitatively different, so comparing overall cost for a desired level of privacy is difficult. Overt transactions are apparent and fingerprintable on the blockchain, but still, provide privacy within the transaction. Disjoint transactions do not link individual users' inputs and outputs on the blockchain. Covert transactions are not fingerprintable as privacy-enhancing transactions.
```
lontivero, line 821:
```suggestion
In \cref{table:coinjoin,table:noncoinjoin}, $n$ denotes the number of participants. The number of transactions is not necessarily comparable because different protocols have significantly different transaction structures. Cost in terms of network fees is better provided in terms of the block weight, in terms of the minimum the number of outputs created and destroyed, with any change outputs incurring additional constant overhead. Overall wait times are also bounded by the minimum number of phases which require transactions to be confirmed in a block. Furthermore, some protocols like Chris Belcher's recent CoinSwap proposal~\cite{belcher2020design} are designed to provide robust privacy with a single iteration of the protocol, whereas others such as Mixcoin~\cite{bonneau2014mixcoin} require multiple mixes to achieve privacy. Secondly, the various protocols are qualitatively different, so comparing overall cost for a desired level of privacy is difficult. Overt transactions are apparent and fingerprintable on the blockchain, but still, provide privacy within the transaction. Disjoint transactions do not link individual users' inputs and outputs on the blockchain. Covert transactions are not fingerprintable as privacy-enhancing transactions.
```

main.tex
nothingmuch: wat, how did that get in there, you're right this is a serious mistake

main.tex
nothingmuch: 64 bit limbs require a 128 bit integer type which is not available in CLR, so even though the hardware is capable of using 64 bit limbs and this is the default for 64 bit builds of libsecp256k1, this is not possible in NBitcoin

main.tex
nothingmuch: ah, i should have read that code more carefully, sorry. I thought "flow" was too vague but i rephrased it to something inaccurate.

main.tex
nothingmuch: hmm, unsure, and i'm not even sure what to google in order to find out the answer

main.tex
nothingmuch: this refers to the last credentials requested, since no registration follows, there is no subsequent request to present them in, it's really just there to make sure the following statement that the sub-graph of interest is k-regular is accurate, since the only way for a directed acyclic graph to truly be k-regular is if it's infinite...

i'm struggling to rephrase it right now, but i will try again a bit later if nobody else suggests anything, it can definitely be clarified

main.tex
MaxHillebrand: I think it's supposed to say `double spending prevention`...

nothingmuch: nope, that's also incorrect.

the only things the credentials ensure is that no unauthorized output registrations are attempted.

if the coordinator is honest, which means only accepts authorized registrations and does not reject any of them, all honest clients should sign the resulting transaction.

all this ensures is that if the coordinator is honest it can still prevent unauthorized registrations, which would either make the transaction invalid or force an honest participant not to sign, and it's only the signature that has anything to do with theft or double spending prevention.

nothingmuch: put another way, if an adversary is able to prove knowledge of a forged MAC, they would be able to register outputs without registering an input or receiving a credential as part of a payment, and an honest coordinator would not be able to output a transaction that all honest users would sign, allowing the adversary to deny service to honest participants indefinitely (i.e. total failure of the protocol).

main.tex
nothingmuch: This is partly redundant with the end of the paragraph, and I think could be a bit clearer overall. How about this?

```suggestion
WabiSabi can be instantiated to construct a variety of CoinJoin transaction structures that depart from the standard output denomination convention, as in SharedCoin\footnote{See: \url{https://github.com/sharedcoin/Sharedcoin}} and CashFusion\footnote{See: \url{https://github.com/cashshuffle/spec}} style transactions and Knapsack~\cite{maurer2017anonymous} mixing. Payments from CoinJoin transactions are possible which enhances sender privacy, as are payments within CoinJoin transactions, improving privacy for both sender and receiver and potentially hiding the payment amount. Both kinds of payments can reduce overall blockspace requirements by reducing the number of intermediate outputs, especially when payments can be batched without compromise privacy. Additionally, restrictions on the consolidation of inputs can be removed and the minimum amount required can be relaxed without degrading the denial of service guarantees of blind signature based protocols.
```

main.tex
nothingmuch: @lontivero - i don't think that redundancy really hurts readability, but it does give another anchor to understand the intuition behind how blind signatures are used so it helps to set up expectations. do you think it's problematic or are you just pointing it out?

with grammar correction:

```suggestion
Several CoinJoin protocols have been implemented and many more proposed, with varying privacy and trust models. For an overview see~\cref{sec:relatedwork}. ZeroLink~\cite{zerolink} is one such protocol implemented by Wasabi, the most popular Chaumian CoinJoin implementation for Bitcoin. Although the coordinator must be trusted to ensure availability, no trust is required with regards to safeguarding users' funds or keeping mapping between inputs and equal amount outputs private. The protocol breaks down into three main phases:
```

main.tex
nothingmuch: The precise behaviour is already described in `sec:limitations` so i think we can just simlify:

```suggestion
We define \emph{CoinJoin inefficiency} as the fraction of non-mixed change outputs in a CoinJoin transaction. In~\cref{fig:cjinefficiency}, we observe the large amount of non-mixed outputs of essentially all Bitcoin privacy-enhancing tools. Samourai has the lowest CoinJoin inefficency due to the homogeneous transaction structure. We note a sharp decline of CoinJoin inefficiency in Wasabi CoinJoins around January 2019 due to the introduction of multiples of the base denomination.
```

main.tex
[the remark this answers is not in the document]
nothingmuch:
```suggestion
To prevent the coordinator from being able to distinguish between initial vs. subsequent input registration requests (which may merge amounts) credential presentation should be mandatory. Initial credentials can be obtained with an auxiliary bootstrapping operation (\cref{fig:bootstrap}). A more sophisticated approach would be to prove either knowledge of a valid MAC or that the amount attribute has a value of zero, but because the input registration phase starts synchronously when enough users have indicated their intent to participate this would not actually save a round trip in practice.
```

main.tex
nothingmuch:
```suggestion
For each amount $a_i$ she also computes a proof that the amounts are in the allowed range, which ensures that finite field arithmetic will not overflow and the values will behave like positive integers when added or subtracted:
```

main.tex
nothingmuch: is this enough?

```suggestion
We can model registration requests as vertices of a directed acyclic graph labeled by $\Delta_a$ and credentials as edges labelled by the amount in the attribute, connecting the registration request where a credential was issued to where it was presented, much like the graph depicted in \cref{sec:examples}. Apart from the bootstrap requests which have indegree 0 (no credentials are presented) and final output requests which have outdegree 0 (credentials must be requested but not used since there are no subsequent registrations), this graph is $k$-regular. The serial number and balance proofs enforce a global invariant where the sums of the inwards and outwards edges of every vertex differ by its label, and so long as honest participants are able to make their registrations it will be balanced (both vertex and edge labels will sum to 0).
```

main.tex
nothingmuch:
```suggestion
The unlinkability of credentials and hiding property of the amount commitments obscure the edges and their labels from the coordinator and other participants. However, the coordinator observes registration requests according to a partial order which is an extension of the partial order defined by the transitive completion of the graph. In other words, the coordinator knows that parallel requests do not share an edge and that dependent requests must be made sequentially. Information about the order and timing of requests as well as the inherently overt transaction data relayed in the registration requests can constrain the graph topology allowing the coordinator to draw inferences about the edges despite not being able to observe the edge set directly.
```

main.tex
nothingmuch:
```suggestion
credential scheme. The denial of service guarantees of the WabiSabi protocol follow from the unforgeability of the KVAC scheme. We refer the to~\cite{chase2019signal} for formal security arguments.
```

main.tex
lontivero: I don't think it is problematic.

main.tex
lontivero: Yes.

main.tex
```diff
@@ -578,105 +732,156 @@ \subsubsection{Malicious Coordinator}
...
+
+\subsection{Theft prevention}
+
+Since the output of the protocol is a single Bitcoin transaction theft is prevented by only signing the transaction if the outputs are as expected, so the protocol inherits theft resistance directly from Bitcoin's security model.
+
+A malicious user could claim more funds than she registered if and only if she is able to forge the coordinator's MAC on some credential. This is certainly not possible unless credentials can be forged under chosen message attacks, but even if that were the case this would only result in denial of service, as the honest users would refuse to sign such a transaction.
+
+\subsection{Security Proofs}
+The above outlined security and privacy guarantees directly follow from the applied KVAC scheme of~\cite{chase2019signal}. The unlinkability of WabiSabi credentials follow from the unlinkability of the underlying keyed-verification anonymous
+credential scheme. The denial of service resistance of the WabiSabi protocol follows from the unforgeability of the KVAC scheme. We refer the to~\cite{chase2019signal} for formal security arguments.
```
nothingmuch, line 779:
```suggestion
The above outlined security and privacy guarantees directly follow from the underlying KVAC scheme of~\cite{chase2019signal}. The unlinkability of WabiSabi credentials follows from the unlinkability and the denial of service resistance of the WabiSabi protocol follows from unforgeability. We refer the to~\cite{chase2019signal} for formal security arguments.
```

main.tex
seresistvanandras: If we have the raw data files, then this is not that much of a work. I can do it real quick :P

seresistvanandras:
Finally had the time to go through the changes! Looking good to me! Let's drop it on e-print, @nothingmuch?

seresistvanandras (approved):

main.tex
nothingmuch: i forgot to take care of this before it was merged, addressed in #99

main.tex
DanGould: I try to wholly avoid the word "mix" because it's so ambiguous to many on-chain practices including some unscrupulous custodial services. Perhaps un-joined. Or standard use. or common-input-heuristic-vulnerable. or "cluster vulnerable".

"Non-fungible" is making it's way into the vernacular and also describes the un-mixed property more concretely as it relates to what the mixing actually does. I like that the best.

p.s. not sure how I managed to lose this mention for so long

PR #98, opened 2021-02-04
Add reissuance request

nopara73:
This PR adds a new type of reissuance request (it's just a fake request for illustration that there's such a thing.)

Note, I did not update the svg as I'm not sure if this is correct.


diagrams/interaction_diagram.txt
```diff
@@ -81,6 +81,15 @@ note right of Coordinator
 end note
 return RealCreds[], ZeroCreds[]
 
+Random Identity -> Coordinator ++: POST credential-reissuance: RoundId, RealCredReqs[], ZeroCredReqs[]
```
nothingmuch, line 84: `ZeroCredReqs[]` and `ZeroCreds[]` can be removed.

If no real credentials were issued yet the real credential requests can only sum to 0, which is a no-op: before and after reissuing the client only has zero credentials and it has the same number before and after issuing the request.

Therefore, it only makes sense to reissue something after connection confirmation has started, at which point real credential issuance is a certainty, and there's no need for the fallback behavior of input registration.

nopara73: That doesn't seem to be right to me or I don't understand how this'd work. See this example: I. should work fine, but II. becomes impossible:

```
I. Assume Alice wants to break 1 coin into 3 coins:

1. input-reg: 2 zerocredreq -> 2 zerocred: 														Alice has 2 zero cred.
2. conn-conf: 2 zero cred pres + 2 real cred req + 2 zero cred req -> 2 real cred, 2 zero cred:	Alice has 2 zero cred and 2 real cred
3. cred-reis: 1 real and 1 zero cred pres + 2 real cred req -> 2 real cred:						Alice has 1 zero cred and 3 real cred

II. Assume Alice wants to break 1 coin into 5 coins: Same things happen, but more cred-reis:
4. cred-reis: 1 real and 1 zero cred pres + 2 real cred req -> 2 real cred:						Alice has 0 zero cred and 4 real cred

Now, we're out of null credentials so we cannot break down more to acquire the 5th credential for the 5th output!
```

nothingmuch: In a discussion with @lontivero yesterday we tried to come up with a simple way to ensure two things:

1. input registration, output registration and reissuance have exactly the same credential validation and issuance
2. number of zero credentials is not limiting (possibly also allowing users without inputs to obtain them and generate reissuance cover traffic)

In my review comment I was tacitly assuming (2) was already solved, but of course this is not true in the current state of the document, sorry.

The approaches we considered were an additional endpoint, which is not too terrible but our least favorite, and making input registration also work for the case of 0 inputs (with no need to confirm anything afterwards).

I need to think more carefully about this approach of always requesting/issuing extra zero credentials, but I think I retract my comment and your approach is indeed simpler.

There's a few other questions that may affect things that I've been trying to work through off github since you asked me to keep it simple, and I don't want to overload with seemingly irrelevant things until I was sure they mattered

nopara73: Got it. Fixing it.

nopara73:
image

diagrams/interaction_diagram.txt
```diff
@@ -1,16 +1,16 @@
...
-collections Satoshis
+collections Anonymous
 participant Coordinator
 participant Alice1
 participant Alice2
 participant Bob
 database Mempool
 
-title Scenario: A Participant Is Consolidating 2 Coins (Alice1, Alice2) Into 1 (Bob)\n
+title Scenario: A Participant Is Consolidating 2 Coins (Alice1, Alice2) Into 1 (Bob) with a fake reissuance request \n
```
MaxHillebrand, line 9: is `fake` the proper word here?

MaxHillebrand:
```suggestion
title Scenario: A Participant Is Consolidating 2 Coins (Alice1, Alice2) Into 1 (Bob) with a unnecessary reissuance request \n
```

diagrams/interaction_diagram.txt
```diff
@@ -6,7 +6,7 @@ participant Alice2
 participant Bob
 database Mempool
 
-title Scenario: A Participant Is Consolidating 2 Coins (Alice1, Alice2) Into 1 (Bob)\n
```
mhmda1, line 9: sequenceDiagram
    participant P as Participant
    participant C as Coordinator
    participant D as Database/Mempool

    title Scenario: Consolidating 2 Coins with Reissuance

    == Input Registration ==
    P->>C: POST credential-reissuance: RoundId, RealCredReqs[], ZeroCredReqs[]
    
    activate C
    Note right of C: 1. IsRoundFound()<br/>2. IsInputConnectionConfirmationPhase()<br/>3. EnsureUniqueSerialNumber()<br/>4. CheckCredProofs()
    
    C-->>P: return RealCreds[], ZeroCreds[]
    deactivate C

    == Output Registration ==
    P->>C: POST output-registration: RoundId, Output, PresentedCreds[]
    activate C
    C-->>P: Success
    deactivate C

mhmda1 (approved):

PR #99, opened 2021-02-25
clarify benchmark results

nothingmuch:
see https://github.com/zkSNACKs/WabiSabi/pull/95#discussion_r570440486


seresistvanandras (approved):
Thanks, Yuval. Looking good to me. Also the language is more formal this way! So kudos to you! :)

PR #107, opened 2022-10-15
Update CoinJoin to coinjoin in this repo

BTCparadigm:
For consistency, shall we also use `coinjoin` instead of `CoinJoin` in this repo?


nopara73:
I personally prefer internet to InterNet.

